PR #2, opened 2018-02-12
Backend setup

teddylo57:


src/main/java/poker/Api.java
```diff
@@ -0,0 +1,32 @@
...
+public class Api {
+
+  public static void main(String[] args) {
+    HandService handService = new HandService();
+    Gson gson = new Gson();
+
+    org.apache.log4j.BasicConfigurator.configure();
+
+    // posts hand data
+    post("/hand", "application/json", (request, response) -> {
```
ericliu121187: There's no required format for the JSON structure right now as we're saving it as a JSON blob in the db right now, correct?

teddylo57: right now there is no required format, it is saving it as it comes in. it can change anytime and does not have to be uniform with the rest of the entries. (as of right now)

PR #3, opened 2018-02-13
Backend setup

teddylo57:


backend/src/main/java/poker/enums/RANK.java
```diff
@@ -0,0 +1,19 @@
+package poker.enums;
+
+public enum RANK {
```
ericliu121187: Can we remove 2s - 5s as per rules?

https://www.pokernews.com/poker-rules/six-plus-hold-em.htm

backend/src/main/java/poker/Api.java
```diff
@@ -0,0 +1,32 @@
...
+public class Api {
+
+  public static void main(String[] args) {
+    HandService handService = new HandService();
+    Gson gson = new Gson();
+
+    //TODO: create actual log4j properties file
+    org.apache.log4j.BasicConfigurator.configure();
+
+    post("/hand", "application/json", (request, response) -> {
```
ericliu121187: In REST-ful convention, can we change the endpoints to be plural?

> 3. Use plural nouns. - https://blog.mwaysolutions.com/2014/06/05/10-best-practices-for-better-restful-api/

> The resource should always be plural in the API endpoint and if we want to access one instance of the resource, we can always pass the id in the URL. - https://hackernoon.com/restful-api-designing-guidelines-the-best-practices-60e1d954e7c9

backend/src/main/java/poker/Api.java
```diff
@@ -0,0 +1,32 @@
...
+      response.type("application/json");
+      Hand hand = gson.fromJson(request.body(), Hand.class);
+      response.status(201); // 201 Created
```
ericliu121187: Didn't say this last time, appreciate the specific `201 Created` response here. Specific codes allow the API users to better debug and handle various scenarios. 🙏 🙏

backend/src/main/java/poker/enums/HANDRANK.java
```diff
@@ -0,0 +1,14 @@
+package poker.enums;
+
+public enum HANDRANK {
```
ericliu121187: If we're going to be using the ordinals, can we update the hand rankings?

https://www.pokernews.com/poker-rules/six-plus-hold-em.htm

backend/src/main/java/poker/enums/RANK.java
teddylo57: Yep

backend/src/main/java/poker/enums/HANDRANK.java
teddylo57: updated

PR #7, opened 2018-02-15
add deck + dealing cards endpoints

teddylo57:
-add deck, cards, hand models
-add shuffle, dealing and show cards endpoints


backend/src/main/java/poker/Api.java
```diff
@@ -23,11 +25,26 @@ public static void main(String[] args) {
       return handService.save(hand);
     });
 
-    get("/hands", (request, response) -> {
+    get("/hands/all", (request, response) -> {
```
ericliu121187: If this is meant to get all the hands, this should probably be `"/hands"` for REST-fulness.

[Rails has a solid basic resources/endpoints convention.](http://guides.rubyonrails.org/routing.html)

Also, for real-life example -- [Mailchimp isn't too bad.](http://developer.mailchimp.com/documentation/mailchimp/reference/overview/)

backend/src/main/java/poker/Api.java
```diff
@@ -23,11 +25,26 @@ public static void main(String[] args) {
...
       response.type("application/json");
       return gson.toJson(handService.list());
     });
 
+    get("/shuffle", (request, response) -> {
```
ericliu121187: When would the client call this endpoint? Irregardless, this would be cool to have to check out shuffling of the deck.

backend/src/main/java/poker/Api.java
teddylo57: good point

backend/src/main/java/poker/Api.java
teddylo57: I suppose it can be built into the dealing of the cards and doesn't have to be exposed for the client. I used it mostly for debugging to ensure cards were shuffled correctly

ericliu121187: Yup

PR #8, opened 2018-02-16
Backend evaluator

teddylo57:


backend/src/main/java/poker/Api.java
```diff
@@ -25,7 +25,7 @@ public static void main(String[] args) {
       return handService.save(hand);
     });
 
-    get("/hands", (request, response) -> {
+    get("/hands/all", (request, response) -> {
```
ericliu121187: Remove `/all`?

backend/src/main/java/poker/Evaluator.java
```diff
@@ -0,0 +1,156 @@
...
+      ranks[i] = 0;
+    }
+    for (Card c : cardList) {
+      ranks[c.getRank().getValue()]++;
+    }
+  //    print();
+    checker(cardList);
+  }
+
+  private void checker(List<Card> cardList) {
```
ericliu121187: Let's go over this together some time so I can better understand

backend/src/main/java/poker/enums/RANK.java
```diff
@@ -1,13 +1,33 @@
...
+  Rank(Integer value) {
+    this.value = value;
+  }
+
+  public Integer getValue() {
+    return value;
+  }
+
+  public static Rank findByValue(Integer value) {
+    for (Rank rank : values()) {
```
ericliu121187: How come we don't use a map here?

backend/src/test/java/poker/DeckTest.java
```diff
@@ -0,0 +1,35 @@
...
+  }
+
+  @Test
+  public void showTest() {
+
+  }
+
+  @Test
+  public void emptyDeck() {
+    for (int i = 0; i < 36; i++) {
```
ericliu121187: I'd avoid using [magic numbers](http://wiki.c2.com/?MagicNumber). Instead, let's declare a constant `"TOTAL_CARD_COUNT"` or something and then set `36` to that.

backend/src/test/java/poker/EvaluatorTest.java
```diff
@@ -0,0 +1,145 @@
...
+  @Test
+  public void checkOnePair() {
+    List<Card> cardList = new ArrayList<>();
+    cardList.add(new Card(Suit.CLOVER, Rank.J));
+    cardList.add(new Card(Suit.SPADE, Rank.J));
+    cardList.add(new Card(Suit.CLOVER, Rank.A));
+    cardList.add(new Card(Suit.CLOVER, Rank.SEVEN));
+    cardList.add(new Card(Suit.CLOVER, Rank.SIX));
+    Evaluator evaluator = new Evaluator(cardList);
+    assertThat("High Card", evaluator.getResult().equals(Handrank.ONE_PAIR));
```
ericliu121187: You have `"High Card"` for all of these tests. That should be different for each test, right?

teddylo57: Yes I have to fix that

backend/src/test/java/poker/DeckTest.java
teddylo57: Agreed it should be extracted out into a variable

backend/src/main/java/poker/enums/RANK.java
teddylo57: We could but that would be an extra step, as seen here..you'd put them into a Map and then retrieve them, we can save that step in my opinion

https://stackoverflow.com/questions/5021246/conveniently-map-between-enum-and-int-string

backend/src/main/java/poker/Evaluator.java
teddylo57: Sure, the code logic will change as I've been working on finding the winning hand for 5-7 cards now.

backend/src/main/java/poker/Api.java
teddylo57: Agreed, not sure why that got merged in

PR #13, opened 2018-02-26
poker hand evaluator

teddylo57:
-add poker hand evaluator
-checks for what hand you possess
-also added comparison to rank which has the highest hand 
-added tests


backend/src/main/java/poker/utils/Evaluator.java
```diff
@@ -0,0 +1,227 @@
...
+          flushSuit = entry.getKey();
+        }
+      }
+    }
+    setFlushSuit();
+
+    if (result.equals(Handrank.FLUSH)) {
+      tieBreakAmount(5);
+    } else if (result.equals(Handrank.STRAIGHT_FLUSH) || result.equals(Handrank.ROYAL_FLUSH)) {
+      tieBreakAmount(1);
```
ericliu121187: We don't need `tieBreakAmount(1)` for `Handrank.STRAIGHT_FLUSH`, right?

Also, how come some places are using `tieBreakers.add(0, tempHigh)` and other places use `tieBreakAmount`? It seems like those are different things.

backend/src/main/java/poker/utils/Evaluator.java
```diff
@@ -0,0 +1,227 @@
...
+
+  private void setPairTieBreakers() {
+    for (int i = ranks.length - 1; i >= 0; i--) {
+      if (ranks[i] == 1 && tieBreakers.size() < 5) {
+        tieBreakers.add(i);
+      }
+    }
+  }
+
+  private void tieBreakAmount(int amount) {
```
ericliu121187: Is this function supposed to just remove elements in the tieBreakers array? If so, why do we need this function, rather than just using like `.clear()`

backend/src/main/java/poker/utils/Evaluator.java
```diff
@@ -0,0 +1,227 @@
...
+    }
+  }
+
+  private void initSuits(Card... cards) {
+    for (Card card : cards) {
+      suitMap.merge(card.getSuit(), 1, (v, vv) -> ++v);
+    }
+  }
+
+  private void checkHand() {
```
ericliu121187: If we check the `result` after each method call, we could prevent additional method calls from happening if unnecessary?

backend/src/main/java/poker/utils/Evaluator.java
teddylo57: The `tiebreakers.add(0, tempHigh)` will add the tiebreaker to the top of the list array, and then the `tieBreakAmount` will see what type of hand you have and remove all elements that doesn't help tiebreak. 

For example: A flush needs 5 cards to tie break, you go from highest to lowest in terms of rank.
A straight flush however only needs one card, the highest card of the straight. Same with a regular straight

backend/src/main/java/poker/utils/Evaluator.java
teddylo57: If you're asking why not clear the arraylist first then add the relevant tiebreakers, we could probably do that instead of overwriting and then clearing the trailing elements

backend/src/main/java/poker/utils/Evaluator.java
teddylo57: Yes, I think that's something I wanted to add, to break out of checking the lower ranked hands in case Flush, Straight Flush or Royal Flush are found. 

4 of a kind does beat a flush but I realized that You can't have 4 of a kind and a Flush in the same hand.

backend/src/main/java/poker/utils/Evaluator.java
teddylo57: Also in the case of a flush of 7 elements (14, 12, 10, 9, 8, 7, 6). This is a Ace high Flush, but it's also a 10 high straight flush. 

We will know it's a flush first. But then we need to check if it's a straight flush. Therefore we we still need to keep elements (14, 12, 10, 9, 8) in case it is not a straight flush. 
Either way, we would need to keep the elements in a temp array or rewrite them later after we check for a Straight Flush/Royal Flush.
I think the method is an easy way to do both. Let me know if you have another solution.

backend/src/main/java/poker/utils/Evaluator.java
ericliu121187: oo yeah good point with the 4 of a kind! yeah there's a lot of logic poker and i think we can leave these optimizations out for now for sure -- although maybe a comment might be nice so we know where to optimize later like a TODO?

backend/src/main/java/poker/utils/Evaluator.java
ericliu121187: most of these thoughts are more related to holding maintainability and readability above speed so keep that in mind, these are my 2c and not saying that this is the best way to do this --

yeah okay so I think this convo is related the other other topic we're chatting about regarding `checkHand`... The concept of symmetry and what a function does is really coming to mind here. 

**What the f is symmetry**
We have `checkStraightFlush` and `checkStraightFlush` methods great! Except they are tucked away somewhere in the depths of `checkFlush` and we're not sure how and where they are being called!

```
// complete psuedocode obvi
// not sure if Java has a switch true statement, which is just a cleaner way writing mutiple if-then statements imo

switch (true) {
  case checkRoyalFlush():
    break;

  case checkStraightFlush():
    break;
  
  case checkFlush():
    break;
```
 
**But eric, you're wasting a lot of time checking for a flush and a straight in checkStraightFlush, but then we have to do it again if it's not a straight flush**
Yeah, we are wasting time to check, but that's all good. Sacrificing a few milliseconds and there and speeding up dev time imo is worth. That's not even to say we can't optimize it later on for faster checks when that becomes an issue.

**Where does this leave us?**
I'm cool with shipping as is if maybe we can name the functions something more descriptive of what they do? `checkFlush` -> `checkFlushHands` or something and the same for the other functions. A little odd that checkFlush leads into that checkStraightFlush and checkRoyalFlush, when Flush StraightFlush and RoyalFlush are all hand ranks. Does that make sense?

**How can we do tiebreaker?**
I kinda sorta get how we're doing tiebreakers here, it def seems like we can can just re-org the 5-card, or 6-card, or 7-card hand, such that we can just go down the elements in the cardlist (and yeah account for like houses or two pairs where there are multiple cards that are in a tiebreak) but that seems like a major re-haul.

if this way works, then it's fine and i suppose the way that we did it here makes the comparator's job easier

**moving forward**
1. As mentioned above, for low hanging fruit, can we move the logic of checkStraightFlush/checkRoyalFlush, etc... up to the checkHand function? If not, let's just rename them to `checkFlushHands` `checkPairHands`, etc.. or something so we know it's doing a level of abstraction above just checking for the fundamental poker hand ranks 
2. Let's keep this way of doing the tiebreakers but is there a way we can make it a little easier to grok? maybe the naming or re-organizing of logic?

teddylo57: Sure, I understand. I can rewrite the code to make it more readable with less confusing method within method calls.
Will look into tiebreakers and how that logic could be easier to understand

backend/src/main/java/poker/utils/Evaluator.java
```diff
@@ -0,0 +1,261 @@
...
+      for (int i = cardList.size() - 1; i >= 0; i--) {
+        if (cardList.get(i).getSuit().equals(flushSuit)) {
+          tieBreakers.add(cardList.get(i).getRank().getValue());
+        }
+      }
+    }
+    return result.equals(Handrank.FLUSH);
+  }
+
+  private boolean aceToNineStraightFlush() {
```
ericliu121187: Don't think this func is accurately named? And kinda violating the single responsibility principle if we're using tieBreakers for this?

Same goes for using the tieBreakers elsewhere for anything other than tieBreaking?

backend/src/main/java/poker/utils/Evaluator.java
```diff
@@ -0,0 +1,261 @@
...
+    return (ranks[14] > 0 &&
+        ranks[6] > 0 &&
+        ranks[7] > 0 &&
+        ranks[8] > 0 &&
+        ranks[9] > 0);
+  }
+
+  private void checkRoyalFlush() {
+    if (checkFlush()) {
+      checkStraightFlush();
```
ericliu121187: Daisy chaining?

backend/src/main/java/poker/utils/Evaluator.java
```diff
@@ -0,0 +1,261 @@
...
+    }
+  }
+
+  private void checkHand() {
+    checkRoyalFlush();
+    checkFlush();
+    checkStraight();
+    checkPairs();
+    setTieBreakers();
+  }
```
ericliu121187: ^ The functions still aren't symmetrical which is okay i guess if `checkPairs` was named more appropriately?

backend/src/main/java/poker/utils/Evaluator.java
teddylo57: I removed the function

backend/src/main/java/poker/utils/Evaluator.java
teddylo57: Removed it and just added the code instead

backend/src/main/java/poker/utils/Evaluator.java
teddylo57: Renamed the function

PR #16, opened 2018-02-28
Handle CORS options request

ericliu121187:
https://github.com/github/fetch/issues/143 -- So when dealing with cross-origin `POST` requests and setting a `content-type` in the header, the client first sends over an `OPTIONS` request expected. 

This didn't occur during using insomnia because we probably never set the content type for the header, though we could have.

This PR handles the CORS options request.


ericliu121187:
Yassssss

Btw got it from http://sparkjava.com/tutorials/cors

PR #18, opened 2018-03-06
refactor code

teddylo57:
-added break/return if a hand is found from ascending order (optimization to not check lower hands)
-added more symmetrical method calls for checking hands
-added some functional programming for streaming through maps


backend/src/main/java/poker/utils/Evaluator.java
```diff
@@ -130,33 +134,44 @@ private void checkForStraight() {
...
-        setIfLargerThanResult(Handrank.FLUSH);
-        flushSuit = entry.getKey();
-      }
+
+  private boolean checkForFlush() {
+    flushSuit = suitMap.entrySet().stream()
+        .filter(map -> map.getValue() >= HAND_SIZE)
+        .map(Map.Entry::getKey)
+        .findFirst()
+        .orElse(null);
```
ericliu121187: Can you explain to me how this stream works? It seems like you're mapping through to get the suit values that are greater than the `HAND_SIZE`? I know I'm missing something here, because I saw your function work.

backend/src/main/java/poker/utils/Evaluator.java
```diff
@@ -130,33 +134,44 @@ private void checkForStraight() {
...
+      return false;
     }
+    setIfLargerThanResult(Handrank.FLUSH);
 
-    if (result.equals(Handrank.FLUSH)) {
-      for (int i = cardList.size() - 1; i >= 0; i--) {
-        if (cardList.get(i).getSuit().equals(flushSuit)) {
-          tieBreakers.add(cardList.get(i).getRank().getValue());
-        }
+    cardList.forEach(s -> {
```
ericliu121187: syntax might be off here, but hope you get the idea -- think it'd be a little nicer if we

```java
flushRanks = cardList.stream()
  .filter(card => card.get().equals(flushSuit))
  .map(card => card.getRank().getValue());
```

Also, why do you use `.equals` rather than `==`?

backend/src/main/java/poker/utils/Evaluator.java
```diff
@@ -58,16 +65,18 @@ private void initSuits(Card... cards) {
   }
 
   private void deterimeHandRank() {
-    checkForRoyalFlushOrStraightFlushOrFlush();
-    checkForStraight();
-    checkForPairs();
-    setTieBreakers();
+    if (checkForStraightFlush()) {
+    } else if (checkForAnyPairs() && checkForFourOfAKind()) {
```
ericliu121187: I think we can "hide" the `checkForAnyPairs` under `checkForFourOfAKind` like we did for `checkForFlush`. Would be nice!

backend/src/main/java/poker/utils/Evaluator.java
```diff
@@ -130,33 +134,44 @@ private void checkForStraight() {
         high = 9;
       }
     }
-    if (straight && setIfLargerThanResult(Handrank.STRAIGHT)) {
-      tieBreakers.add(0, high);
-    }
+    return straight && setIfLargerThanResult(Handrank.STRAIGHT);
```
ericliu121187: In JS, our functions often just return a falsy value even though an operation has been performed (usually `undefined`). In Java, is it common to return `true` explicitly like you did in `setIfLargerThanResult`?

backend/src/main/java/poker/utils/Evaluator.java
```diff
@@ -81,35 +90,30 @@ private void checkForPairs() {
...
-      }
-    } else if (pair1 == 2 && pair2 == 1) {
-      if (setIfLargerThanResult(Handrank.ONE_PAIR)) {
-        tieBreakers.add(0, largeGroupRank);
-      }
-    } else if (pair1 == 1) {
-      setIfLargerThanResult(Handrank.HIGH_CARD);
-    }
+  private boolean checkForFullHouse() {
+    return pair1 == 3 && pair2 == 2 && setIfLargerThanResult(Handrank.FULL_HOUSE);
```
ericliu121187: Wondering if `setIfLargerThanResult` could have a better name (maybe I came up with that one haha) like `updateResultIfNecessary`? `setResultIfNeeded`? `setResultIfNecessary`? I think "larger" just seems a little off  and what are we setting...

backend/src/main/java/poker/utils/Evaluator.java
```diff
@@ -58,16 +65,18 @@ private void initSuits(Card... cards) {
   }
 
   private void deterimeHandRank() {
-    checkForRoyalFlushOrStraightFlushOrFlush();
-    checkForStraight();
-    checkForPairs();
-    setTieBreakers();
+    if (checkForStraightFlush()) {
```
ericliu121187: Didn't catch this before, I'm thinking we should rename this function to `determineHandRank` because English?

backend/src/main/java/poker/utils/Evaluator.java
```diff
@@ -130,33 +134,44 @@ private void checkForStraight() {
...
 
     int count = 0;
-    int tempHigh = 0;
-    for (int i = 0; i < tieBreakers.size() - 1; i++) {
-      if (tieBreakers.get(i) - 1 == tieBreakers.get(i + 1)) {
-        if (tieBreakers.get(i) > tempHigh) {
-          tempHigh = tieBreakers.get(i);
+    tempHigh = 0;
+    for (int i = 0; i < flushRanks.size() - 1; i++) {
+      if (flushRanks.get(i) - 1 == flushRanks.get(i + 1)) {
```
ericliu121187: How come we have to make this seemingly odd check here? We're checking to see if the current value - 1 is equal to the next value?

backend/src/main/java/poker/utils/Evaluator.java
ericliu121187: Ohhh, is it because we already have a suitMap that has all the suits mapped and the corresponding list of cards in that suit? And then we're checking which suit has more than 5 cards? If that's the case, then question -- where are mapping to suitMap? I don't see where we set that those values.

backend/src/main/java/poker/utils/Evaluator.java
teddylo57: oops typo

backend/src/main/java/poker/utils/Evaluator.java
teddylo57: sure, was trying to not hide anything like last time 😰

backend/src/main/java/poker/utils/Evaluator.java
teddylo57: I guess in this case, it'll check for a straight and only return true if it is a straight and the result wasn't all the hand ranks above it... although I guess it would never execute this code if the result has been set 🤔

backend/src/main/java/poker/utils/Evaluator.java
teddylo57: initSuits method whenever a card is added to the class

backend/src/main/java/poker/utils/Evaluator.java
teddylo57: https://stackoverflow.com/questions/7520432/what-is-the-difference-between-vs-equals-in-java

backend/src/main/java/poker/utils/Evaluator.java
teddylo57: hm it's the straight checking logic, for example since sorted highest to lowest it'll be 10,9,8,7,6. To compare 10 to 9, it'll check 10 - 1 == 9? And if there are 4 consecutive matches it should be a straight

backend/src/main/java/poker/utils/Evaluator.java
```diff
@@ -170,17 +185,66 @@ private void checkForRoyalFlushOrStraightFlushOrFlush() {
...
     }
-    if (result.equals(Handrank.STRAIGHT_FLUSH)) {
-      tieBreakers.add(0, tempHigh);
+    return (result.equals(Handrank.ROYAL_FLUSH) || result.equals(Handrank.STRAIGHT_FLUSH));
+  }
+
+  private void setTieBreakers() {
+    switch (result) {
+      case ROYAL_FLUSH:
+        tieBreakers.add(0, tempHigh);
```
ericliu121187: How come in some of these cases, you're adding an index? or a number? before the value? And other ones like `tieBreakerers.add(largeGroupRank)`, you aren't?

backend/src/main/java/poker/utils/Evaluator.java
```diff
@@ -170,17 +185,66 @@ private void checkForRoyalFlushOrStraightFlushOrFlush() {
...
+        elementsNeededForTiebreak(1);
+        break;
+      case STRAIGHT_FLUSH:
+        tieBreakers.add(0, tempHigh);
+        elementsNeededForTiebreak(1);
```
ericliu121187: I wonder if the concept of `elementsNeededForTiebreak` is necessary. Can we just add the necessary value(s) to the tieBreakers array? 

For example, for straight flush, all we need is the `tempHigh`. For full house, we put in `largeGroupRank` and `smallGroupRank`. For two pair, we put in the higher of pair, then lower pair, and then the last card. For high card, we put in all 5 cards sorted.

teddylo57: Might be some edge cases like a 7 card flush and only taking the 5 highest cards but I can take a look once I land

backend/src/main/java/poker/utils/Evaluator.java
ericliu121187: oo, does it make sense though why we are okay with doing it here and for the checkFlush inside checkForStraightFlush? At least, now for maintainability we know that the high flush hands (straight flush) already do the checking of a flush for us, and then for the highest paired hands (four of a kind) already do the checking of pairs for us, so the logic is symmetrical in a sense

backend/src/main/java/poker/utils/Evaluator.java
```diff
@@ -170,17 +185,66 @@ private void checkForRoyalFlushOrStraightFlushOrFlush() {
...
-        tieBreakers.contains(9) &&
-        tempHigh != 0) {
-      result = Handrank.STRAIGHT_FLUSH;
+    if (flushRanks.contains(14) &&
+        flushRanks.contains(6) &&
+        flushRanks.contains(7) &&
+        flushRanks.contains(8) &&
+        flushRanks.contains(9) &&
+        9 >= tempHigh) {
+      setIfLargerThanResult(Handrank.STRAIGHT_FLUSH);
```
ericliu121187: Do we ever `setIfLargerThanResult(Handrank.STRAIGHT_FLUSH);` if it's a "regular" straight flush?

backend/src/main/java/poker/utils/Evaluator.java
ericliu121187: Dang, I kinda get it now. Would be so nice if we had that checkForStraight function handle this kind of logic too, no biggie.

teddylo57: Wrote a todo to write a data structure (combine flushRanks and ranks) that can be passed to a straight method

backend/src/main/java/poker/utils/Evaluator.java
ericliu121187: what do you think of the syntax i suggested?

teddylo57: sure, added

PR #19, opened 2018-03-07
Websocket chat

teddylo57:


backend/pom.xml
```diff
@@ -43,7 +43,17 @@
     <dependency>
       <groupId>com.sparkjava</groupId>
       <artifactId>spark-core</artifactId>
-      <version>2.7.1</version>
+      <version>2.7.2</version>
+    </dependency>
+    <dependency>
+      <groupId>com.j2html</groupId>
```
ericliu121187: Do we need this?

teddylo57: not sure if you need it to render html, if not i can delete

backend/src/main/java/poker/Api.java
```diff
@@ -14,6 +15,8 @@
 public class Api {
 
   public static void main(String[] args) {
+    webSocket("/chat", ChatWebSocketHandler.class);
```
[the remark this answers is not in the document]
teddylo57: yes it works, but i am unable to e2e test without the corresponding js component

backend/src/main/java/poker/models/Table.java
```diff
@@ -0,0 +1,8 @@
+package poker.models;
+
+import java.util.List;
+
+public class Table {
```
ericliu121187: Is this all that is needed in the Table class? Does this handle seating and standing for players?

teddylo57: added seating and players leaving the table

backend/src/main/java/poker/models/Table.java
```diff
@@ -0,0 +1,28 @@
...
+
+  public void addPlayer(Player newPlayer) throws Exception{
+    if (players.size() < 8) {
+      players.add(newPlayer);
+    } else {
+      throw new Exception("Table is full");
+    }
+  }
+
+  public void removePlayer(Player deletePlayer) {
```
ericliu121187: is there a better name than `deletePlayer`? that sounds a lot like a function name

PR #20, opened 2018-03-08
Table

teddylo57:


backend/pom.xml
```diff
@@ -43,7 +43,17 @@
     <dependency>
       <groupId>com.sparkjava</groupId>
       <artifactId>spark-core</artifactId>
-      <version>2.7.1</version>
+      <version>2.7.2</version>
```
ericliu121187: Some of these code changes seem unnecessary

backend/src/main/java/poker/models/DealtHand.java
```diff
@@ -4,11 +4,11 @@
 
 @Data
 public class DealtHand {
+
```
ericliu121187: i 👀u bro

backend/src/main/java/poker/models/Player.java
```diff
@@ -0,0 +1,24 @@
+package poker.models;
+
+import lombok.Data;
+
+@Data
+public class Player {
+  private String uuid;
```
ericliu121187: nit - new line above

backend/pom.xml
teddylo57: Actually spark 2.7.1 had a bug that threw a logging error because it depended on a jetty websocket dependency that was updated in 2.7.2

backend/pom.xml
```diff
@@ -43,7 +43,17 @@
...
+    </dependency>
+    <dependency>
+      <groupId>com.j2html</groupId>
```
ericliu121187: Are `j2html` and `json` necessary?

backend/pom.xml
ericliu121187: Was this what you needed to upgrade to get it working?

backend/src/main/java/poker/services/HandService.java
```diff
@@ -4,7 +4,7 @@
 import java.util.List;
 import org.mongodb.morphia.Datastore;
 import org.mongodb.morphia.Morphia;
-import poker.models.dao.Hand;
+import poker.models.dao.HandHistory;
 
 public class HandService {
```
ericliu121187: Probably want to rename this `HandHistoryService`

backend/src/test/java/poker/HandTest.java
```diff
@@ -0,0 +1,88 @@
...
+import poker.models.Deck;
+import poker.models.Hand;
+import poker.models.Player;
+
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.equalTo;
+import static org.hamcrest.Matchers.hasItem;
+import static org.hamcrest.Matchers.not;
+
+public class HandTest {
```
ericliu121187: Can we make the styling in this file consistent at least within itself?

backend/src/main/java/poker/models/DealtHand.java
```diff
@@ -7,8 +7,7 @@
   private Card card1;
   private Card card2;
 
-  public DealtHand(){
-    this.card1 = Deck.getInstance().deal();
-    this.card2 = Deck.getInstance().deal();
+  public DealtHand() {
```
ericliu121187: Do we still need this class?

backend/src/main/java/poker/models/Hand.java
```diff
@@ -0,0 +1,64 @@
...
+
+  public void removePlayer(Player leavingPlayer) {
+    players.remove(leavingPlayer);
+  }
+
+  public int getPlayerSize() {
+    return players.size();
+  }
+
+  public void dealStartingHand() {
```
ericliu121187: `dealStartingHands` or `dealHands`?

ericliu121187:
> That being said, I'd prefer getting code in and if you make the changes here, let's 🚢 ! I'll create tickets for those.

Sorry, to clarify "make the changes here", I meant "make the changes previously mentioned in the comments", not including the "I'm thinking like a table can have an array of tournaments. A tournaments can have an array of hands. Let's also keep in mind we want to keep the concept of hand abstracted in a way that is tournament/cash game agnostic."

That way we could have gotten that code in, my bad!

backend/pom.xml
teddylo57: No, upgrading the version helped get rid of a logging error

backend/src/main/java/poker/models/DealtHand.java
ericliu121187: why

teddylo57: I think it's useful to group the two cards dealt to a player in a class because it would be cleaner to inject into the DB as a class rather than card1 and card2 for each player and also readability. I think it's like factoring a polynomial

backend/src/main/java/poker/models/Player.java
```diff
@@ -0,0 +1,25 @@
+package poker.models;
+
+import lombok.Data;
+
+@Data
+public class Player {
+
+  private String uuid;
+  private String name;
```
ericliu121187: Change to `username`

backend/src/main/java/poker/models/Player.java
```diff
@@ -0,0 +1,25 @@
+package poker.models;
+
+import lombok.Data;
+
+@Data
+public class Player {
+
+  private String uuid;
```
ericliu121187: Remove this for now

backend/pom.xml
```diff
@@ -43,7 +43,12 @@
     <dependency>
       <groupId>com.sparkjava</groupId>
       <artifactId>spark-core</artifactId>
-      <version>2.7.1</version>
+      <version>2.7.2</version>
+    </dependency>
+    <dependency>
+      <groupId>org.json</groupId>
```
ericliu121187: remove until we need?

PR #21, opened 2018-03-09
Created a table that has 8 seats with sit button

txu89:
screen shot 2018-03-09 at 2 01 58 am

This PR has two components.
* The child component renders the seat with a sit button. The logic for sitting has not been implemented yet.
* The parent components displays the poker table container. It loops the seat component 8 times.


frontend/src/components/PokerTable.js
```diff
@@ -0,0 +1,37 @@
+/* @flow */
+import React, { Component } from 'react';
+
+const tableStyle = {
+  border: '1px solid blue',
+};
+
+export class PokerTable extends Component {
```
ericliu121187: Make stateless component

frontend/src/components/PokerTable.js
```diff
@@ -0,0 +1,37 @@
...
+
+    for (let i = 0; i < NUM_OF_SEATS; i++) {
+      seats.push(<TableSeat />);
+    }
+
+    return <div style={tableStyle}>{seats}</div>;
+  }
+}
+
+class TableSeat extends Component {
```
ericliu121187: Make functional component

ericliu121187:
The poker table should only be rendered for logged in user. I believe we spoke about [lifting state up](https://reactjs.org/docs/lifting-state-up.html).

> The logic for sitting has not been implemented yet.

Just confirming that this is preferred since it's in another ticket. 👍

frontend/src/components/PokerTable.js
```diff
@@ -0,0 +1,37 @@
...
+  border: '1px solid blue',
+};
+
+export class PokerTable extends Component {
+  constructor(props) {
+    super(props);
+  }
+
+  render() {
+    let seats = [];
```
ericliu121187: Does seats need to be a `let` here?

txu89:
Before Login:
screen shot 2018-03-09 at 9 35 49 pm
After Login:
screen shot 2018-03-09 at 9 35 59 pm

State has been lifted from LoginField component into App.
The table also does not show until after logging in.

The border style is temporary to visualize seats/table.

frontend/src/components/App.js
```diff
@@ -1,17 +1,51 @@
...
+        <header>
+          <h1>{'Short Deck Poker'}</h1>
+        </header>
+        <LoginField
+          isLoggedIn={isLoggedIn}
+          onLoginSubmit={this.handleLoginSubmit}
+          onUsernameChange={this.handleUsernameChange}
+          username={username}
+        />
+        <PokerTable isLoggedIn={isLoggedIn} />
```
ericliu121187: Instead of passing this as a prop, simply render it based on `isLoggedin`.

```javascript
{isLoggedIn && <PokerTable />}
```

frontend/src/components/LoginField.js
```diff
@@ -5,37 +5,29 @@ export class LoginField extends Component {
...
-    this.onUsernameChange = this.onUsernameChange.bind(this);
-    this.state = {
-      isLoggedIn: false,
-      username: '',
-    };
+    this.handleChange = this.handleChange.bind(this);
   }
 
   handleLoginSubmit(e) {
+    const { onLoginSubmit } = this.props;
```
ericliu121187: Prefer `e.preventDefault();` to be the first thing in this method. By convention since the dawn of JS, `e.preventDefault();` goes first in the function. Just convention.

```javascript
e.preventDefault();
const { onLoginSubmit } = this.props;
onLoginSubmit();
```

frontend/src/components/PokerTable.js
```diff
@@ -0,0 +1,31 @@
+/* @flow */
+import React from 'react';
+
+//temporary styling to see TableSeat
```
ericliu121187: Nice comment 👍 nit: I've rarely, if ever, seen no space between `//` and the comment

```javascript
// temporary styling to see TableSeat
```

frontend/src/components/PokerTable.js
```diff
@@ -0,0 +1,31 @@
...
+const tableStyle = {
+  border: '1px solid blue',
+};
+
+export function PokerTable(props) {
```
ericliu121187: 3 things --

#### 1. Let's `export` this as `default`.
It's common for files that are named after a component to export the component as a default.

```javascript
export default function PokerTable(props)
```

#### 2. Almost always the default export is at the very bottom of the file
Move this below `<TableSeat />`.

#### 3. This is 100% style, but let's try this out!
Destructure props in the parameter list rather than in the function body.

```javascript
export function PokerTable({ isLoggedIn }) {
```

^ Will be gone though since we don't need this prop

frontend/src/components/PokerTable.js
```diff
@@ -0,0 +1,31 @@
...
+  const NUM_OF_SEATS = 8;
+
+  for (let i = 0; i < NUM_OF_SEATS; i++) {
+    seats.push(<TableSeat />);
+  }
+
+  if (isLoggedIn) {
+    return <div style={tableStyle}>{seats}</div>;
+  } else {
+    return null;
```
ericliu121187: You can set a guard if `!isLoggedIn` to `return null;`, it'll prevent declaring the variables and running through the loop.

```javascript
const { isLoggedIn } = props;

if (!isLoggedIn) {
  return nulll;
}

const seats = [];
const NUM_OF_SEATS = 8;

for (let i = 0; i < NUM_OF_SEATS; i++) {
  seats.push(<TableSeat />);
}

return <div style={tableStyle}>{seats}</div>;
```

PR #23, opened 2018-03-11
Establish basic Websocket connection 

ericliu121187:
https://trello.com/c/mvhyARQS/174-8-user-can-connect-via-websocket

⚠️ Make sure you're running `Api.java` and the FE app with `yarn start`.

#### What we have here

This PR establishes a very basic connection joining the BE and FE. This is a pretty much a POC. I left comments of where we will need to do work. I prefaced the comments with OOS, meaning out-of-scope for this PR.

#### It works!

For now this is the success flow.
1. User types in username (not really important for this).
1. User hits login, which triggers us attempting to connect through websocket.
1. If successful, there's an alert.

#### How you can tell it doesn't work

In the console of the FE upon typing a username and hitting the login button, if you get the following error, you know there's a problem.

#### Concerns

~@teddylo57 I kept getting these NPEs while loading up the api. How can we fix? Should we be concerned?~ Bumping up `spark-core` to 2.7.2 fixed this


teddylo57:
Update to spark 2.7.2

backend/src/main/java/poker/Api.java
```diff
@@ -1,22 +1,23 @@
...
+import static spark.Spark.*;
+
 public class Api {
 
   public static void main(String[] args) {
+
     HandService handService = new HandService();
     Gson gson = new Gson();
 
+    webSocket("/connect", UsersWebsocketHandler.class); // must come before CorsFilter.apply()
```
ericliu121187: I tried putting this line in several places. I thought it should be with the paths below, since it's basically a path too, but it was not working when I put it after `CorsFilter.apply()`. Not entirely sure of where to put this line... so here it is.

backend/src/main/java/poker/Users.java
```diff
@@ -0,0 +1,27 @@
...
+import org.json.JSONObject;
+
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+
+public class Users {
+
+    // this map is shared between sessions and users, so it needs to be thread-safe (http://stackoverflow.com/a/2688817)
+    static Map<Session, String> userUsernameMap = new ConcurrentHashMap<>();
+    static int nextUserNumber = 1; // TODO Remove and use username
```
ericliu121187: OOS Because we don't parse any messages that the FE sends over yet, we don't have a way of communicating what the FE is sending over as the username that the user is attempting to login with. For now, we'll just use this basic `nextUserNumber` and record users as `User1`, `User2`, etc...

frontend/src/components/App.js
```diff
@@ -15,7 +15,26 @@ class App extends Component {
...
+    webSocket.onmessage = function(msg) {
+      console.log(msg);
+    };
+  }
+
   handleLoginSubmit() {
+    const { username } = this.state;
+    const PORT = 4567; // TODO use env variable
+    const webSocket = new WebSocket(
+      `ws://${window.location.hostname}:${PORT}/connect` // TODO move to constant
```
ericliu121187: OOS A bunch of TODOs here -- we need to clean up these variables and start using an `.env` file. Upcoming!

ericliu121187:
@teddylo57 Yup, that fixed it

teddylo57:
For the Java styling, we can use 
https://github.com/google/styleguide

Import the `intellij-java-google-style.xml` into IntelliJ

backend/src/main/java/poker/Api.java
```diff
@@ -11,59 +11,59 @@
...
-
-    get("/cards", (request, response) -> {
-      response.type("application/json");
-      return gson.toJson(Deck.getInstance().getRemainingCards());
-    });
-
-  }
-}
\ No newline at end of file
+    public static void main(String[] args) {
```
teddylo57: I think you formatted it to IntelliJ default rather than google style, you probably didn't import it into the IDE 

https://github.com/HPI-Information-Systems/Metanome/wiki/Installing-the-google-styleguide-settings-in-intellij-and-eclipse

teddylo57: I think Java style for Google is 2 tabs vs 4 tabs for intelliJ default

PR #24, opened 2018-03-11
fix recompose

txu89:


ericliu121187:
This is needed because I accidentally installed the recompose lib from Mark43 repo from my work comp

PR #25, opened 2018-03-12
Add username validation and /login route

teddylo57:
- Added a new route /login to validate username by JSON request.
- Added username validation of the request JSON ->
**Rules are:**
A user can only login if the username is valid
No other users are connected with the same username.
A username may only contain alphanumeric characters so no underscores, dashes, or periods.
A username has to be between 3-20 characters, inclusive.
A username may not start with a number.
Two usernames are considered the same case-insensitive. 

Sample JSON:
```
{
  username: "ericliu121187"
}

### CODE REVIEW

File: backend/pom.xml [hunk 1]
```diff
@@ -86,5 +86,10 @@
       <artifactId>slf4j-log4j12</artifactId>
       <version>1.7.5</version>
     </dependency>
+    <dependency>
```
teddylo57: Dependency added for string validations

backend/src/main/java/poker/Api.java
```diff
@@ -43,10 +44,21 @@ public static void main(String[] args) {
...
-      response.status(201); // 201 Created
+      response.status(201);
       return handHistoryService.save(handHistory);
     });
 
+    post("/login", "application/json", (request, response) -> {
+      response.type("application/json");
+      Player player = gson.fromJson(request.body(), Player.class);
+      if (!Users.isValidUsername(player.getUsername())) {
+        response.status(409);
```
teddylo57: 409 Conflict response

backend/src/main/java/poker/Api.java
```diff
@@ -43,10 +44,21 @@ public static void main(String[] args) {
...
+        return "User is attempting to connect with an invalid username.";
+      }
+      response.status(201);
```
ericliu121187: For this case, we can use either the `200 OK` or `202 Accepted`? `201 Created` is used usually for situations when we create a new record in the db.

I've usually seen `200 OK` for successful logins, but maybe this is a good scenario to use `202 Accepted`, like we accept your `username` and now it's up to the FE to connect and finish the login process.

backend/src/main/java/poker/Api.java
ericliu121187: I'm thinking we use `409 Conflict response` for when there is a conflict in uniqueness of the username. Let's use `401 Unauthorized` here.

https://stackoverflow.com/questions/11714485/restful-login-failure-return-401-or-custom-response

backend/src/main/java/poker/Api.java
teddylo57: Sure, that makes sense. Changed to `202`

backend/src/main/java/poker/Users.java
```diff
@@ -23,4 +24,14 @@ public static void broadcastMessage(String sender, String message) {
       }
     });
   }
+
+  public static boolean isValidUsername(String username) {
+    return !userUsernameMap.values().contains(username) &&
```
ericliu121187: Not sure if you saw the OOS tag on the card, but can we address this separately `!userUsernameMap.values().contains(username)`? Because technically it's a valid username, but there is a conflict with the state of the server and the FE and user should be aware of it.

teddylo57: Removed the two checks for conflict in the username map

PR #28, opened 2018-03-13
Check for unique username during POST to /login

teddylo57:
Checks if the username is unique within connected websocket sessions, if so, the server will respond with a 202 and a JSON including the player name. From here I think the FE can hit the websocket endpoint `/connect` and then the backend can associate a `Session` with the `Username` sent in the `POST` request to `/login` and the FE can transition the page to the `Table`

-Add methods to check if the username entered is unique and not taken within a current active websocket session
-If unique, returns 202
-If username is taken, returns 409 with a "Username is already taken" response


backend/src/main/java/poker/Api.java
```diff
@@ -15,6 +15,8 @@
 
 public class Api {
 
+  static String username;
```
teddylo57: Used to persist the username that is sent on the POST request to /login

ericliu121187: Would this work if there are two players? Would this `username` get overridden?

backend/src/main/java/poker/Api.java
```diff
@@ -55,7 +57,12 @@ public static void main(String[] args) {
         response.status(401);
         return "User is attempting to connect with an invalid username.";
       }
+      if (!Users.isAvailable(player.getUsername())) {
+        response.status(409);
+        return "Username is already taken";
```
ericliu121187: nit: Put a period at the end for consistency w/ all other messages

backend/src/main/java/poker/Users.java
```diff
@@ -31,4 +30,10 @@ public static boolean isValidUsername(String username) {
         !Character.isDigit(username.charAt(0));
   }
 
+  public static boolean isAvailable(String username) {
+    return !userUsernameMap.values().contains(username) &&
```
ericliu121187: Do we need `!userUsernameMap.values().contains(username)` ?

backend/src/main/java/poker/Api.java
teddylo57: I think its fine as it should be thread safe, will test once FE updates code to initiate websocket collection if received a 202

backend/src/main/java/poker/Users.java
```diff
@@ -31,4 +30,10 @@ public static boolean isValidUsername(String username) {
...
+        userUsernameMap.values().stream().map(String::toLowerCase)
+            .noneMatch(user -> user.equals(username.toLowerCase()));
```
ericliu121187: If you don't want to recalc `username.toLowerCase()` every time, we can move it out of the stream. Something like:

```java
String lowerCaseUsername = username.toLowerCase();
return userUsernameMap.values().stream().map(String::toLowerCase)
    .noneMatch(user -> user.equals(lowerCaseUserName));
```

Side q: does Google handle the styling for chaining things, i see 4 spaces and i see this in general are there special rules for indentation for chaining?

backend/src/main/java/poker/Api.java
ericliu121187: Are there diff instances of this API for each player?

Player1 joins. `username` is "Player1". 

Player2 join. `username` is "Player2".

Player1 leaves. ```Users.broadcastMessage(sender = "Server", msg = (Api.username + " left the table."));``` <-- what will that broadcast?

Taking a step back, maybe we want to implement a `Session` class? You're right that we should store the `username`, but maybe not in the `Api` class?

teddylo57: Hm, I can get the username from the Map based on their Session for the `Users.broadcastMessage` method

Sure, I was thinking of refactoring the `Api` class, it's a bit crowded and should be split up into different functions

teddylo57: I can work on it in another ticket/PR since it wouldn't fit into this scope of `unique username` validation. Thoughts?

ericliu121187: Works for me 👍

PR #30, opened 2018-03-13
POST to /login with username

ericliu121187:
### Changes
When user attempts to login, we now hit `/login`. Depending on if it's a valid username name or not, we get back the appropriate response. <- Nice work @teddylo57 

Pretty cool, make sure you update your config (see below) and if you run both the FE and BE, you should see the responses in the Network tab.

### Upcoming
A couple things -- 
1. The BE needs to return a JSON object. Right now we're getting back strings. No bueno.
1. We need to finish the flows.

I'll create tickets.

### Config
⚠️ You need to create a `.env` file in your root `frontend/`. It should look something like the following. You can change the `REACT_APP_DEV_API_URL` to whatever port you're running `Api.java` from.

```
NODE_PATH=src/
REACT_APP_DEV_API_URL=http://localhost:4567
```


frontend/src/components/PokerTable.js
```diff
@@ -38,6 +38,7 @@ const TableSeatWrapper = styled.div`
         return 'transform: rotate(190deg) translate(26rem) rotate(-190deg);';
       case 8:
         return 'transform: rotate(216deg) translate(20rem) rotate(-216deg);';
+      // no default
```
ericliu121187: This was needed for https://eslint.org/docs/rules/default-case

frontend/.flowconfig
```diff
@@ -8,3 +8,4 @@ src/flow/types.js
 [lints]
 
 [options]
+module.name_mapper='^constants\(\/?.*\)$' -> '<PROJECT_ROOT>/src/constants/\1'
```
ericliu121187: Need to tell Flow where to look for imports/exports

frontend/.gitignore
```diff
@@ -19,3 +19,5 @@
 npm-debug.log*
 yarn-debug.log*
 yarn-error.log*
+
+.env
```
ericliu121187: Don't wanna have your `.env` exposed on the interwebs

PR #31, opened 2018-03-14
Add Error Response class -> JSON to be returned from /login 

teddylo57:
Return JSON object from /login using an `ErrorResponse` class that can be extended to hold other types of `errors`. Here we just have 
```
{ 
  errors: {
    username: "Here is the error string"
  }
}
```
However we can add other fields besides `username` and return empty fields if null in the future. 
Like 
```
{ 
  errors: {
    authentication: "Here is the error string"
  }
}
```
if username field was null and `authentication` was not.

-Add Error Response class which contains Errors 
-Errors class can have other errors thrown and returned to the front end as an object


backend/src/main/java/poker/models/ErrorResponses.java
```diff
@@ -0,0 +1,16 @@
...
+@Data
+public class ErrorResponses {
+
+  private Errors errors;
+
+  @Data
+  public class Errors {
+
+    private String username;
+
```
teddylo57: Can add other fields in the future for other errors

PR #32, opened 2018-03-14
Refactor Api Class  

teddylo57:
-Refactored the `Api` class into a `Routes` class with methods that can be called such as `userRoutes` and `debugRoutes` for testing. This decouples routes from our `Api` class and even the `Cors` filter stuff.
-Added proper `log4j` logging properties so we can set levels of `DEBUG` or `INFO` in the future.
-Deleted some `Card` routes that were just used for testing and debugging our application during development.
-Rename classes but feel like there are better names, open to suggestions


backend/src/main/java/poker/Routes.java
```diff
@@ -0,0 +1,74 @@
...
+import poker.models.ErrorResponses;
+import poker.models.Player;
+import poker.models.dao.HandHistory;
+import poker.services.handhistory.HandHistoryService;
+import poker.services.users.UserService;
+
+public class Routes {
+
+  private Gson gson = new Gson();
+  public static String username;
```
ericliu121187: I'm still not convinced that this is the best place to put `username`, but we can leave for now.

backend/src/main/java/poker/Routes.java
```diff
@@ -0,0 +1,74 @@
...
+  Routes() {
+    handHistoryRoutes();
+    userRoutes();
+    debugRoutes();
+  }
+
+  private void handHistoryRoutes() {
+    HandHistoryService handHistoryService = new HandHistoryService();
+
+    post("/hands", "application/json", (request, response) -> {
```
ericliu121187: For consistency, change to `/hand_histories`?

backend/src/main/java/poker/Routes.java
```diff
@@ -0,0 +1,74 @@
...
+      response.type("application/json");
+      HandHistory handHistory = gson.fromJson(request.body(), HandHistory.class);
+      response.status(201);
+      return handHistoryService.save(handHistory);
+    });
+
+    get("/hands", (request, response) -> {
```
ericliu121187: .

backend/src/main/java/poker/services/CorsFilter.java
```diff
@@ -32,4 +34,21 @@ public void handle(Request request, Response response) throws Exception {
     };
     Spark.after(filter);
   }
+
+  public static void setCorsHeaders() {
```
ericliu121187: would `setCorsHeadersForOptions` or `handleOptionsRequests` describe this method clearer?

backend/src/main/java/poker/services/users/UserService.java
```diff
@@ -1,17 +1,17 @@
...
 import org.json.JSONObject;
 
-public class Users {
+public class UserService {
 
   // this map is shared between sessions and users, so it needs to be thread-safe (http://stackoverflow.com/a/2688817)
   static Map<Session, String> userUsernameMap = new ConcurrentHashMap<>();
 
-  public static void broadcastMessage(String sender, String message) {
+  public static void sendChatToActiveUsers(String sender, String message) {
```
ericliu121187: interested in your reasoning behind this change

a couple of things --

1. we don't really differentiate between active and non-active users right now, so i don't think that's fair to say?
1. `chat` doesn't capture everything as much as `message` does - bc it's not really a chat, maybe a notification, like notify all users -- that being said, i do kinda lean towards broadcast, it's captures the intent -- other words i would consider are `announce` or `transmit`, both which don't fit perfectly, `announce` -- what if it's not an announcement, `transmit` sounds like radio waves or an out-dated system, also `message` in this case is also a very specific websocket term because in websocket land we send messages back and forth

that being said, i think we can go with:
`sendMessageToAllUsers`
`sendMessageToUsers`
`broadcastMessageToUsers`
`broadcastMessage` <-- maybe this is enough?
`broadcastMessageToAllUsers` <-- t captures that it _is all_ users, if the function name was `broadcastMessage` i might expect it to take a list of users that we want to broadcast to - maybe the `all` is redundant though

good luck -- if it was up to me, i think i wouldn't feel strongly enough to change the original name, and thus, would leave it?

backend/src/main/java/poker/Api.java
```diff
@@ -1,100 +1,20 @@
...
-
   public static void main(String[] args) {
-    HandHistoryService handHistoryService = new HandHistoryService();
-    Gson gson = new Gson();
-
-    webSocket("/connect", UsersWebsocketHandler.class); // must come before CorsFilter.apply()
+    webSocket("/connect", UsersWebsocketHandler.class);
 
-    //TODO: look into CORS filter before a release
+    //TODO: look into CORS stuff before a release
```
ericliu121187: is there something you think we need to look into for CORS? else, i think we can remove this line

backend/src/main/java/poker/Routes.java
teddylo57: I agree, I can put a TODO and we can revisit it

backend/src/main/java/poker/services/users/UserService.java
teddylo57: Ok will leave it

backend/src/main/java/poker/Routes.java
ericliu121187: something like
```java
public static String username; // probably will want to move elsewhere
```
?

teddylo57: I made it a //TODO: revisit where to place

teddylo57: IntelliJ should let us view all TODO's
https://www.jetbrains.com/help/idea/todo-example.html
I guess we can revisit before a release?

ericliu121187: sweet

PR #33, opened 2018-03-16
Table sitting permissions

teddylo57:
When a player wants to sit at the table, he will click a button with the `seatNumber` to sit down. This code will allow a player to sit and will lock that seat so another player cannot sit down in that seat.

I created a class `Request` that will have `data` and `type`. I guess based on the `type` of action that is passed from the front end, I can determine what logic to perform. I'm not sure if we will always be using the `/connect` websocket or we will initiate more for say `/chat`. Which is why I implemented  the request to act as a switch on the type, we can call a `chat` method if the `type : "CHAT"`. 

I also moved around some classes as some were related to `websocket` models and some were `gameplay` models.


backend/src/main/java/poker/Api.java
```diff
@@ -2,11 +2,14 @@
 
 import static spark.Spark.webSocket;
 
+import poker.models.game.Table;
 import poker.services.CorsFilter;
 import poker.services.users.UsersWebsocketHandler;
 
 public class Api {
 
+  public static Table testTable = new Table("SDP 10 ETH tournament");
```
teddylo57: Our test table

teddylo57:
A good site for testing websocket stuff
https://www.websocket.org/echo.html

backend/src/main/java/poker/models/game/Table.java
```diff
@@ -0,0 +1,31 @@
...
+
+  private String name;
+  private boolean[] availableSeats;
+
+  public Table(String name) {
+    this.name = name;
+    initalizeSeats();
+  }
+
+  public boolean isAvailable(int seatNumber) {
```
ericliu121187: `isSeatAvailable`?

backend/src/main/java/poker/models/websocket/Request.java
```diff
@@ -0,0 +1,13 @@
...
+
+import lombok.Data;
+import poker.enums.seat.Sit;
+import poker.models.game.Seat;
+
+@Data
+public class Request {
+
+  private Seat data;
+  private Sit type;
```
ericliu121187: Prob refactor this soon, `Seat` and `Sit` is too specific

backend/src/main/java/poker/models/game/Table.java
```diff
@@ -0,0 +1,31 @@
...
+import java.util.Arrays;
+import lombok.Data;
+
+@Data
+public class Table {
+
+  private static final int MAX_TABLE_SIZE = 8;
+
+  private String name;
+  private boolean[] availableSeats;
```
ericliu121187: Instead of boolean, can we actually include the User that is seated? Or rather, where are we storing the data of where the user is seated?

backend/src/main/java/poker/enums/seat/Sit.java
```diff
@@ -0,0 +1,7 @@
+package poker.enums.seat;
+
+public enum Sit {
```
ericliu121187: Yeah i'd make this something like MessageTypes, can do this later or I guess MessageType since all of the other enums are singular

backend/src/main/java/poker/services/users/UsersWebsocketHandler.java
```diff
@@ -1,33 +1,46 @@
...
-    UserService.broadcastMessage(sender = "Server", msg = (Routes.username + " joined the table."));
+    UserService.broadcastMessage(sender = "[SERVER] ", msg = (Routes.username + " joined the table."));
   }
 
   @OnWebSocketClose
   public void onClose(Session user, int statusCode, String reason) {
     String leavingUsername = UserService.userUsernameMap.remove(user);
     UserService.userUsernameMap.remove(user);
-    UserService.broadcastMessage(sender = "Server",
+    UserService.broadcastMessage(sender = "SERVER] ",
```
ericliu121187: typo

backend/src/main/java/poker/models/websocket/Request.java
teddylo57: Agreed, we should create a generalized model for `Request` and `Responses` from the websocket so it can decode and encode respectively

ericliu121187:
> I'm not sure if we will always be using the /connect websocket or we will initiate more for say /chat.

I think we talked about this in person, but just confirming -- once we are connected, we'll be communicating via websocket messages, so no we don't need /chat. We need to make some changes to the messaging architecture soon, i'll update the tech doc.

teddylo57:
FYI: I tested by hardcoding `username` in `Routes` to equal "tedder"

backend/src/main/java/poker/services/users/UserService.java
```diff
@@ -24,6 +31,31 @@ public static void broadcastMessage(String sender, String message) {
...
+    Response response = new Response();
+    Response.Data sitResponseData = response.new Data();
+
+    if (Api.testTable.isSeatAvailable(seatNumber)) {
+      Api.testTable.sitDown(seatNumber, userUsernameMap.get(user));
+      sitResponseData.setSuccess(true);
+      sitResponseData.setRequest(request);
+      response.setType(MessageType.SIT_DOWN_SUCCESS);
+    } else {
+//      Api.testTable.sitDown(seatNumber);
```
ericliu121187: remove

PR #34, opened 2018-03-16
Set up FE redux and epics; Send message through Websocket

ericliu121187:
[CHECK IT OUT](https://cl.ly/2e2d2v0C2g2a)

### Overview
Most of this PR is adding new packages and setting up our app to use [Redux](https://redux.js.org/) -- this follows [the FE architecture outlined here](https://github.com/ericliu121187/short-deck-poker/tree/master/frontend#folder-structure). I need to make some updates to that architecture which will be coming up shortly along with updates to the main README.

Ask me anything!

### What changed
The flow for the FE now goes like:
1. user writes username
1. user hits login
1. we POST /login
1. if valid, we initiate websocket connection and send over the username

^ there's more work to come such as handling of errors from /login

### ❗️What you need to do
1. Install redux dev tools. I set it up to integrate with that, if you look at my demo above, you'll notice a tab on my dev tools named "Redux" the same should happen for you and you can play around with it. Easier to debug now -- https://chrome.google.com/webstore/detail/redux-devtools/lmhkpmbekcpmknklioeibfkpmmfibljd?hl=en 
1. Add `REACT_APP_DEV_WEBSOCKET_URL=ws://localhost:4567` to your `.env` or wherever your websocket url is.


frontend/src/constants/messageTypeEnum.js
```diff
@@ -0,0 +1,10 @@
+/* @flow */
+import keyMirror from 'keymirror';
+
+const messageTypeEnum = keyMirror({
```
ericliu121187: @teddylo57 on the FE, we'll just have a `messageTypeEnum` I think it might make sense to have the same on BE

teddylo57: Agreed

frontend/src/features/session/index.js
```diff
@@ -0,0 +1,150 @@
...
+);
+
+export const usernameSelector = createSelector(
+  baseSessionSelector,
+  session => session.username
+);
+
+// EPICS
+
+const loginEpic = (action$, { getState }) =>
```
ericliu121187: This epic is pretty cool -- we look for `LOGIN_KICKOFF` then we look for `LOGIN_SUCCESS` and then we `CONNECT_KICKOFF` before throwing back `LOGIN_SUCCESS`. 💥 

We use reactive programming to handle our side effects. What are they? 

Basically we have a stream of observables. These observables have side effects like... `LOGIN_KICKOFF` kicks off a call to `/login`. So we use epics to handle them.

https://en.wikipedia.org/wiki/Side_effect_(computer_science)
https://gist.github.com/staltz/868e7e9bc2a7b8c1f754 <-- good easy intro article to reactive programming which is what we use here

frontend/src/app/index.js
```diff
@@ -0,0 +1,13 @@
...
+import { createEpicMiddleware } from 'redux-observable';
+import rootEpic from './rootEpic';
+import rootReducer from './rootReducer';
+import { applyMiddleware, createStore } from 'redux';
+
+const epicMiddleware = createEpicMiddleware(rootEpic);
+
+export default createStore(
+  rootReducer,
+  window.__REDUX_DEVTOOLS_EXTENSION__ && window.__REDUX_DEVTOOLS_EXTENSION__(),
```
teddylo57: How come it seems like this is called twice, once as a parameter and once as a method?

ericliu121187: It's not called twice. In JS, you invoke a function with the `()` as you prob expected. The `&&` evaluates it as a expression. So what this does it basically

Hey! Check if `window.__REDUX_DEVTOOLS_EXTENSION__` exists (if it's truthy) and if it exists, call it!

This prevents us from calling a function that doesn't exist, because in this particular case, not everyone has downloaded that `REDUX_DEVTOOL_EXTENSION`. So if they haven't, don't call it bc it'll throw an error.

frontend/src/features/session/index.js
teddylo57: Can you detail the full `MessageType` back and forth conversation?

You send a `CONNECT_KICKOFF` and then look for a `LOGIN_SUCCESS` from the back end?

teddylo57: screen shot 2018-03-18 at 1 20 50 am

I see this is the first message sent through the websocket

ericliu121187: > Can you detail the full MessageType back and forth conversation?

[Did so in this card](https://trello.com/c/P4aAh2qp/196-2-handle-attempt-to-connect), is there anything I'm missing?

teddylo57: -After the `202` if the username is valid the backend returns back just the username
-Front end initiates a websocket connection.
-First message sent is in the format `{ data: {username: eric}, type: }` with `CONNECT_KICKOFF`
-Backend responds with `LOGIN_SUCCESS` or `LOGIN_FAILURE` ?
And that completes the flow for login?

teddylo57: Ok so the FE's first message of sending `CONNECT_KICKOFF` through the websocket - does the backend have to respond to complete the flow?

ericliu121187: per convo, let's prefix `WS_` to message types through websocket

PR #37, opened 2018-03-17
User is able to click on Seat Number button to sit down at table

txu89:
screen shot 2018-03-17 at 2 17 26 am

* Added a new state `takenSeat` to `PokerTable`
* Pass `takenSeat`, `handleSit`, and `username` as props to `TableSeat`

Will conditionally render the username after seat button is clicked.


frontend/src/components/PokerTable.js
```diff
@@ -44,21 +45,62 @@ const TableSeatWrapper = styled.div`
...
+    );
   });
 
   return <PokerTableWrapper>{seats}</PokerTableWrapper>;
 }
+
+export default compose(
+  withState('takenSeat', 'setTakenSeat', -1),
+  withHandlers({
+    handleSit: ({ setTakenSeat }) => (seatNum: number): void => {
```
ericliu121187: it's good to have the same naming convention within a file, or even a codebase ofc

in this case, what i am suggesting is `seatNum` -> `seatNumber`

codebases might use `btn` or `button`, but try to be consistent

frontend/src/components/PokerTable.js
```diff
@@ -44,21 +45,62 @@ const TableSeatWrapper = styled.div`
...
+        username={username}
+      />
+    );
   });
 
   return <PokerTableWrapper>{seats}</PokerTableWrapper>;
 }
+
+export default compose(
+  withState('takenSeat', 'setTakenSeat', -1),
```
ericliu121187: i _like_ the usage of `-1` here, i def see this a lot more in static languages as it needs to be the same type and if it needs to be a number it can use `-1` to represent a falsy value

in JS i've seen it a little less, but no need to change here -- this also makes our flow type nice bc we use type `number` and we don't have to worry about `undefined` or `null`

frontend/src/components/PokerTable.js
```diff
@@ -44,21 +45,62 @@ const TableSeatWrapper = styled.div`
...
+  number,
+  takenSeat,
+  username,
+}: {
+  onSit: number => void,
+  number: number,
+  takenSeat: number,
+  username: string,
+}) {
+  if (number !== takenSeat) {
```
ericliu121187: in programming there is this concept of single responsibility principle so this concept of TableSeat should be only responsible for a table seat... let's think about it for a second, if you were a table seat at a table, do you _really_ care about where other people are seated? as in do you _care_ if `takenSeat` is your seat? here you're having the table seat needing to know where the person is seated

as a diff approach, what might be slightly cleaner is if you only care if there was a person seating in you? and if so, who?

https://en.wikipedia.org/wiki/Single_responsibility_principle

no need to change here, just food for thought

txu89: This makes sense. I will take this into account when adding stand up feature to the seats.

PR #38, opened 2018-03-18
Use Player class instead of username

teddylo57:
Instead of passing around the username everywhere, a username belongs to a `Player`. Therefore we can just extract the `username` field from the `Player` when we need it. I thought about if we need a `User` and a `Player` class but think the only difference would be a `Player` would be playing and not spectating which is what a `User` would be. In the future when we allow spectators, perhaps a `Player` can be extended from a User class but for now I don't think it's worth the abstraction. Cards will only be dealt to `Players` that are sitting so we can check for that.


teddylo57:
I went with `Player` instead of `User` because right now we don't make any distinction on the front end to differentiate the two. Unless you are thinking a `User` becomes a `Player` every time they sit down? I do think `Player` extends a `User` class but there is no point in making it right now when the two have the same abilities as of right now

PR #39, opened 2018-03-18
Backend: handle connect attempt 

teddylo57:
Implemented methods to return a successful websocket connection to the FE. Included some enums to make our lives easier in trying to navigate flows through messages received from the `/connect` websocket.

-Create `Action` class to differentiate flow of messages through the websocket connection
-Use a `switch` statement to navigate different flows based on `MessageType` from FE
-Add flow for responding to FE request for a websocket connection


teddylo57:
I tested before adding the `WS_` prefix as discussed but once they are merged on both FE/BE then it should work properly as in the GIFs

teddylo57:
[Success](https://drive.google.com/open?id=1Dg35pKeLajYSEQzYbF-HFmUuRo4mQ1mp)
[Error](https://drive.google.com/open?id=1sUZMxRfLjkYd1hIob6JW64-nVwBuN5_O)
[2 Player Success](https://drive.google.com/open?id=1GMJvQ-526mSIdisj3h4SneDNUPEHuWsa)

backend/src/main/java/poker/services/users/UsersWebsocketHandler.java
```diff
@@ -35,14 +34,20 @@ public void onClose(Session user, int statusCode, String reason) {
   }
 
   @OnWebSocketMessage
-  public void onMessage(Session user, String message) throws IOException {
+  public void onMessage(Session user, String message) throws IOException, Exception {
     Request request = gson.fromJson(message, Request.class);
-
-    for (MessageType messageTypeType : MessageType.values()) {
-      if (request.getType().equals(messageTypeType)) {
+    switch (request.getType().getValue()) {
```
ericliu121187: How come we don't just use `request.getType()` as the expression here?

teddylo57: I suppose we could, if it's always 1:1 from enum to function call. Since I guess the generic pattern will be kickoff followed by success or failure

backend/src/main/java/poker/enums/websocket/MessageType.java
```diff
@@ -1,7 +1,30 @@
 package poker.enums.websocket;
 
 public enum MessageType {
-  SIT_DOWN_KICKOFF,
-  SIT_DOWN_SUCCESS,
-  SIT_DOWN_FAILURE;
+  WS_SIT_DOWN_KICKOFF(Action.SIT),
```
ericliu121187: Why do this instead of just having `WS_SIT_DOWN_KICKOFF`, `WS_SIT_DOWN_SUCCESS`, etc...?

teddylo57: You mean with grouping the enum into actions? I guess we can. I was unclear on the kickoff, success, failure model for most requests. I can just look for kickoff

ericliu121187: i'm just not sure what the `Action` class provides us, are you planning on using it and separating out the `UserService` methods such as `UserService.connectUser(user, request);` and `UserService.sitUser(user, request);`?

ericliu121187:
the gifs are great! so basically, everything is working as expected

backend/src/main/java/poker/enums/websocket/MessageType.java
teddylo57: I can get rid of it. I guess it was before I understood the ticket message type pattern

teddylo57:
Yeah just the FE has to use the WS_ prefix and also only let a user connect via websocket if it gets a 202

backend/src/main/java/poker/enums/websocket/MessageType.java
ericliu121187: Yeah I think we can remove it and just use the enums straight up for now. We can make another level of abstraction if we find that it's too hard to manage the switch case "state machine".

ericliu121187: Also, the FE concept of `Action` is a redux framework concept, not business logic.

ericliu121187: https://redux.js.org/basics/actions

backend/src/main/java/poker/services/users/UserService.java
```diff
@@ -24,14 +25,37 @@ public static void broadcastMessage(String sender, String message) {
...
         e.printStackTrace();
       }
     });
   }
 
+  public static void connectUser(Session user, Request request) throws IOException {
+    Gson gson = new Gson();
+
+    Response response = new Response();
+    Response.Data sitResponseData = response.new Data();
```
ericliu121187: `connectResponseData`?

PR #42, opened 2018-03-20
Feat/stand button

txu89:
screen shot 2018-03-19 at 11 55 01 pm

* Changed the `takenSeat` prop to a boolean `isSeated` for `TableSeat` to let `PokerTable` handle figuring out assigning seats. 
* When seated the empty seats have sit button removed and render `'Empty seat'` instead. This needs to be styled.
* Conditionally render a stand button when seated to remove user from the seat and choose another. Need to style to fix location of button or remove it from `PokerTable` component.


frontend/src/components/PokerTable.js
```diff
@@ -72,28 +72,59 @@ function TableSeat({
...
 
-  times(NUM_OF_SEATS, i => {
-    seats.push(
-      <TableSeat
-        number={i + 1}
-        onSit={handleSit}
-        takenSeat={takenSeat}
-        username={username}
-      />
+  if (takenSeat < 0) {
```
ericliu121187: this works great -- looking over all of this it seems like we can condense this a bit maybe to something like

```javascript
times(NUM_OF_SEATS, i => {
  const seatNumber = i + 1;
  const isSeated = seatNumber === takenSeat;

  seats.push(
    <TableSeat
      isSeated={isSeated}
      number={seatNumber}
      onSit={handleSit}
      username={isSeated ? username : 'Empty Seat'}
  />
});
```

frontend/src/components/PokerTable.js
```diff
@@ -72,28 +72,59 @@ function TableSeat({
...
+            onSit={handleSit}
+            username={`Empty Seat`}
+          />
+        );
+      }
+    });
+    return (
+      <PokerTableWrapper>
+        {seats}
+        <button onClick={handleStand} type="button">{`Stand`}</button>
```
ericliu121187: If we get rid of these if-then statements, we can still conditionally render the stand button by checking to see if takenSeat is a legit seat number.

```javascript
{takenSeat !== -1 && (
  <button onClick={handleStand} type="button">{`Stand`}</button>
)}
```

frontend/src/components/PokerTable.js
```diff
@@ -102,5 +133,8 @@ export default compose(
     handleSit: ({ setTakenSeat }) => (seatNumber: number): void => {
       setTakenSeat(seatNumber);
     },
+    handleStand: ({ setTakenSeat }) => (): void => {
+      setTakenSeat(-1);
```
ericliu121187: maybe at the top of this file we can put something like

```javascript
const STANDING = -1
```

frontend/src/components/PokerTable.js
```diff
@@ -72,28 +72,59 @@ function TableSeat({
...
+            username={username}
+          />
+        );
+      } else {
+        seats.push(
+          <TableSeat
+            isSeated
+            number={i + 1}
+            onSit={handleSit}
+            username={`Empty Seat`}
```
ericliu121187: Seems like this might be a good opportunity to rename `username` prop to something more semantic like `display` or `label` since we're now using it for empty seats which aren't usernames

txu89:
Refactored conditional rendering of `TableSeat` and `PokerTable`.
Also added and changed some variables to improve clarity.

frontend/src/components/PokerTable.js
```diff
@@ -48,52 +48,66 @@ const TableSeatWrapper = styled.div`
...
 }: {
   handleSit: number => void,
+  handleStand: number => void,
   takenSeat: number,
   username: string,
 }) {
-  const seats = [];
   const NUM_OF_SEATS = 8;
+  const seats = [];
+  const STANDING = -1;
```
ericliu121187: This should go at the top of the file so that way we can use it for the `-1` below.

frontend/src/components/PokerTable.js
```diff
@@ -48,52 +48,66 @@ const TableSeatWrapper = styled.div`
...
+  const isSeated = takenSeat > STANDING;
```
ericliu121187: I think we should go `!==`, `>` implies there's some sort of logic in the amount comparison where there shouldn't be.

frontend/src/components/PokerTable.js
```diff
@@ -48,52 +48,66 @@ const TableSeatWrapper = styled.div`
...
+        seatLabel={isUserSeat ? username : 'Empty Seat'}
       />
     );
   });
 
-  return <PokerTableWrapper>{seats}</PokerTableWrapper>;
+  return (
+    <PokerTableWrapper>
+      {seats}
+      {takenSeat !== STANDING && (
```
ericliu121187: We can use the `isSeated` variable here?

txu89:
Moved `STANDING` and fixed variables.

ericliu121187:
> Moved STANDING and fixed variables.

while i appreciate these comments, they aren't necessary if you have good commit messages, which you do 👍 and that's even better

i also get notified every time you push a commit

PR #43, opened 2018-03-21
Start a hand

teddylo57:
`BE` will send `FE` a message to initialize the tournament. 
-Table ID
-Hand ID
-Button
-Seating
-Players (names + chips)


backend/src/main/java/poker/models/game/Hand.java
```diff
@@ -2,18 +2,31 @@
...
 
 @Data
 public class Hand {
 
   private String tableName;
-  private List<Player> players = new ArrayList<>();
+  private int handId;
+  private List<Player> players;
+  private List<Seat> seats;
+  private Player button;
```
ericliu121187: I think this logic should also be in `Table`, alongside the `seats`. Also, what if a player gets moved? Should `button` be of type `Seat`?

backend/src/main/java/poker/models/game/Hand.java
```diff
@@ -2,18 +2,31 @@
...
+import java.util.stream.Collectors;
 import lombok.Data;
 
 @Data
 public class Hand {
 
   private String tableName;
-  private List<Player> players = new ArrayList<>();
+  private int handId;
+  private List<Player> players;
```
ericliu121187: This info seems redundant as we already have this in `seats`. However, I would argue that we should probably remove `seats` and the corresponding methods out of this class and into `Table`. Let's leave this here and then refactor the seats logic out?

backend/src/main/java/poker/models/game/Player.java
```diff
@@ -7,6 +7,7 @@
 
   private String username;
   private DealtHand hand = new DealtHand();
+  private int chips = 20000;
```
ericliu121187: To prevent magic numbers -- can we have something like 

```
final static private int STARTING_CHIPS = 20000;
```

?

backend/src/main/java/poker/models/game/Table.java
```diff
@@ -8,27 +16,56 @@
...
   private String name;
-  private Seat[] seats;
+  private int id;
+  private ArrayList<Seat> seats;
+  private int playerSize;
+  private ArrayList<Hand> hand;
+  private Player button;
 
-  public Table(String name) {
+  public Table(String name, int id) {
```
ericliu121187: I usually see ids being generated from the db. Seeing how we're not writing that to the DB right now, is it common to initialize a class with an id as an argument like this?

backend/src/main/java/poker/models/game/Table.java
```diff
@@ -8,27 +16,56 @@
...
+    playerSize++;
   }
 
   private void initalizeSeats() {
-    this.seats = new Seat[MAX_TABLE_SIZE];
-    for (int i = 0; i < seats.length; i++) {
-      seats[i] = new Seat();
+    this.seats = new ArrayList<>(Arrays.asList(new Seat[MAX_TABLE_SIZE]));
+    for (int i = 0; i < seats.size(); i++) {
+      seats.set(i, new Seat());
```
ericliu121187: Should all these references to `seats` be referring to `this.seats`?

backend/src/main/java/poker/models/game/Hand.java
```diff
@@ -2,18 +2,31 @@
...
+  private List<Seat> seats;
+  private Player button;
   private List<Card> burnList = new ArrayList<>();
   private List<Card> flop = new ArrayList<>();
   private Card turn;
   private Card river;
 
+  public Hand(Table table) {
+    this.tableName = table.getName();
```
ericliu121187: I see you do stuff like `table.getName()` but I don't see these getters being set. Is there a library that handles this for us?

teddylo57: Yes I use lombok as a library that does that for us, classes with the `Data` annotation will have `getters` and `setters` boilerplate injected

https://projectlombok.org/
https://github.com/rzwitserloot/lombok

backend/src/main/java/poker/models/game/Table.java
teddylo57: Yes, I haven't written anything to the DB yet, should I write to the DB without having a hand dealt?

backend/src/main/java/poker/models/game/Player.java
teddylo57: I was drunk

backend/src/main/java/poker/models/game/Hand.java
teddylo57: Sure, will take a look

backend/src/main/java/poker/models/game/Hand.java
teddylo57: Was not sure if button should be a `Seat` or a `Player`. I suppose it should be a `Seat`? 

In a tournament, a `Player` should not get moved right?

backend/src/main/java/poker/models/game/Table.java
ericliu121187: Nah, don't worry about writing to DB, just curious

backend/src/main/java/poker/models/game/Table.java
ericliu121187: Well, besides clarity, doesn't it mean different things or no?

backend/src/main/java/poker/models/game/Hand.java
ericliu121187: You're right for our application and the implementation would work if it's a player, but after giving it more thought, I think it's incorrect to have it type `Player`.

Should be `Seat` because while we don't have this logic yet, players can get up, switch tables, sit down in another table, and the button shouldn't follow them.

teddylo57:
Tested with websocket echo website 

```
{
  "data": {
    "success": true,
    "table": {
      "name": "SDP 10 ETH tournament",
      "id": 1,
      "seats": [
        {
          "seatNumber": 1
        },
        {
          "seatNumber": 2
        },
        {
          "seatNumber": 3
        },
        {
          "player": {
            "hand": {
              "card1": {
                "suit": "SPADE",
                "rank": "EIGHT"
              },
              "card2": {
                "suit": "DIAMOND",
                "rank": "SEVEN"
              }
            },
            "chips": 20000
          },
          "seatNumber": 4
        },
        {
          "player": {
            "hand": {
              "card1": {
                "suit": "CLOVER",
                "rank": "J"
              },
              "card2": {
                "suit": "SPADE",
                "rank": "A"
              }
            },
            "chips": 20000
          },
          "seatNumber": 5
        },
        {
          "seatNumber": 6
        },
        {
          "seatNumber": 7
        },
        {
          "seatNumber": 8
        }
      ],
      "playerSize": 2,
      "hand": {
        "tableName": "SDP 10 ETH tournament",
        "handId": 0,
        "burnList": [],
        "flop": []
      },
      "button": {
        "player": {
          "hand": {
            "card1": {
              "suit": "SPADE",
              "rank": "EIGHT"
            },
            "card2": {
              "suit": "DIAMOND",
              "rank": "SEVEN"
            }
          },
          "chips": 20000
        },
        "seatNumber": 4
      }
    }
  },
  "type": "WS_HAND_START"
}
```

teddylo57:
https://drive.google.com/open?id=1rTfyMAIqxRAafrygxqG3MV_7_-wUQSiq

backend/src/main/java/poker/models/game/Table.java
```diff
@@ -8,27 +15,57 @@
...
+  public void start() {
+    if (button == null) {
+      int randomButton = ThreadLocalRandom.current().nextInt(0, getOccupiedSeats().size());
+      button = getOccupiedSeats().get(randomButton);
+      button = seats.get(randomButton);
+    } else {
+      moveButton();
     }
+    this.hand = new Hand(this);
+    this.hand.dealHands(getOccupiedSeats());
```
ericliu121187: We can leave for now, but I imagine we would want all the head dealing logic to be within `Hand` instead of here.

Would we be able to call methods like `this.table.getButton` and `this.table.getOccupiedSeats` in Hand?

Also, now that I'm thinking about it more. Maybe later on, it's a litter cleaner to have Table with Players, instead of Table with Seats...

ericliu121187:
We can merge this, but regarding the hole cards, that ticket isn't finished as not every player should only see his/her own hole cards.

backend/src/main/java/poker/models/game/Table.java
teddylo57: Sure, we can call those methods as long as they are `public`. 

Sure that makes sense

PR #46, opened 2018-03-24
Blinds

teddylo57:
I added big blinds and small blinds to the table. I'm guessing the next card will be a `MoneyService` or `Bank` in which it actually deducts the sb/bb from users and adds it to a `pot` field for `table`. It should also handle `betting` but I guess we can discuss that before you create a card.

**Main changes:**
I made a constructor for `Deck` because as a singleton wouldn't make sense with multiple tables. Along with the fact that each `hand` will be using a fresh deck, no sense in keeping a shared deck.

Added the table websocket `ws_hand_start` syncs all active players with the starting table information.

Then `ws_deliver_hole_cards` will send each individual player their hole cards. I think it's a good idea, since everything else will be in a shared state with the table. The only unique piece of data would be the hole cards, so it makes sense to send them separately.

I fixed some tests and added some for button and blind logic.

Examples of table data: 
```
{"data":{"success":true,"table":{"name":"SDP 10 ETH tournament","id":1,"seats":[{"seatNumber":1},{"seatNumber":2},{"seatNumber":3},{"player":{"username":"ted","chips":20000},"seatNumber":4},{"player":{"username":"bob","chips":20000},"seatNumber":5},{"seatNumber":6},{"seatNumber":7},{"seatNumber":8}],"playerSize":2,"hand":{"tableName":"SDP 10 ETH tournament","handId":0,"burnList":[],"flop":[]},"button":{"player":{"username":"ted","chips":20000},"seatNumber":4},"blinds":{"smallBlind":{"amount":25,"player":{"username":"ted","chips":20000}},"bigBlind":{"amount":50,"player":{"username":"bob","chips":20000}}},"pot":{"pot":0}}},"type":"WS_HAND_START"}
```

Examples of Card data:
```
{"data":{"success":true,"dealtHand":{"card1":{"suit":"HEART","rank":"J"},"card2":{"suit":"SPADE","rank":"SEVEN"}}},"type":"WS_DELIVER_HOLE_CARDS"}
```

```
{"data":{"success":true,"dealtHand":{"card1":{"suit":"CLOVER","rank":"NINE"},"card2":{"suit":"DIAMOND","rank":"A"}}},"type":"WS_DELIVER_HOLE_CARDS"}
```

I was able to test and verify these JSON websocket messages. I think this takes us a lot closer to being to start a tournament game.


backend/src/main/java/poker/models/game/Hand.java
```diff
@@ -9,45 +9,48 @@
...
-        .forEach((player -> player.setHand(Deck.getInstance().deal())));
+        .forEach((player -> player.setHand(deck.deal())));
     seatedPlayers.stream().map(Seat::getPlayer)
-        .forEach((player -> player.setHand(Deck.getInstance().deal())));
+        .forEach((player -> player.setHand(deck.deal())));
   }
 
   private void burn() {
-    Card burn = Deck.getInstance().deal();
+    Card burn = deck.deal();
```
ericliu121187: don't we need to use `this.deck` in these places? does `this.deck` _and_ `deck` mean the same thing? or can we write `deck` as shorthand?

i thought `deck` means the static field that is shared by all instances of this class whereas `this.deck` refers to the field on a particular instance

backend/src/main/java/poker/models/game/Hand.java
```diff
@@ -9,45 +9,48 @@
 
   private String tableName;
   private int handId;
+  private transient Deck deck;
```
ericliu121187: how come we want `transient` here? i thought that `transient` is used for values/fields that are derived from other fields. for example, if we have seats and we only want the occupied seats, we could have occupied seats as a transient field, because we're deriving that off seats.

backend/src/main/java/poker/services/gameplay/GameService.java
```diff
@@ -25,9 +27,27 @@ public static void startTournament(Session user) throws IOException {
       tableResponseData.setSuccess(true);
       tableResponseData.setTable(Api.testTable);
       response.setData(tableResponseData);
-
       UserService.sendAll(gson.toJson(response));
+      sendHoleCards();
+    }
+  }
+
+  public static void sendHoleCards() {
```
ericliu121187: `deliverHoleCards` as to be in line with the messageType?

backend/src/main/java/poker/models/game/Player.java
```diff
@@ -8,7 +8,7 @@
   final static private int STARTING_CHIPS = 20000;
 
   private String username;
-  private DealtHand hand = new DealtHand();
+  private transient DealtHand hand = new DealtHand();
```
ericliu121187: Maybe it'd be nice in the future if we renamed `DealtHand` to `HoleCards`?

backend/src/main/java/poker/models/game/Table.java
```diff
@@ -44,12 +48,15 @@ public void start() {
...
-      button = seats.get(randomButton);
     } else {
       moveButton();
     }
     this.hand = new Hand(this);
-    this.hand.dealHands(getOccupiedSeats());
+    setBlinds();
+  }
+
+  public void end() {
```
ericliu121187: How come you put this logic here instead of `Hand`?

backend/src/main/java/poker/models/game/Hand.java
teddylo57: We can use `deck` as shorthand here since `deck` and `this.deck` refer to the same field. But we can append `this.` to `deck` to make it clearer.

https://stackoverflow.com/questions/3728062/what-is-the-meaning-of-this-in-java

So I changed `deck` from being a static class to being normal constructor class. Since each hand will have a `new Deck`. Also with different tables, we won't be sharing the same deck with each table. `static` is useful if I'm sharing something across the entire program.

backend/src/main/java/poker/models/game/Hand.java
teddylo57: https://stackoverflow.com/questions/4802887/gson-how-to-exclude-specific-fields-from-serialization-without-annotations

I use transient so it omits the `deck` from being serialized into JSON and sent over the websocket.

backend/src/main/java/poker/models/game/Table.java
teddylo57: I did it to be symmetrical, since `Table` has a `start()` method, it makes sense for the `Table` to also end the hand. Other things that would probably be called in the `end()` method would be calling the `MoneyService` to distribute the pot to the winner/

teddylo57: I suppose if you want to have a `clear()` method in `Hand` and then the `Table` could have an `end()` method that calls `hand.clear()` that would also work.

I guess it depends if you see the `Table` doing the clearing or the `Hand` doing the clearing

PR #49, opened 2018-03-25
Websocket feature; Handle `WS_SIT_DOWN_SUCCESS`, `WS_CONNECT_SUCCESS`

ericliu121187:
I moved the websocket functionality into its own feature folder, handled `WS_SIT_DOWN_SUCCESS` and `WS_CONNECT_SUCCESS`. Regarding the latter, now the login flow does not complete until we receive back a `WS_CONNECT_SUCCESS` so you may notice some delay after you log in. That is expected and correct.

https://trello.com/c/jkYDDliY/211-3-sit-down-epic
https://trello.com/c/GANdloIs/222-2-finish-refactoring-of-login-flow-using-the-websocket-epic


frontend/src/features/websocket/index.js
```diff
@@ -0,0 +1,61 @@
...
+export function sendWebsocketMessage(type: MessageType, data: mixed) {
+  socket$.next(JSON.stringify({ data, type }));
+}
+
+// EPICS
+
+const websocketMessageEpic = action$ =>
+  action$.ofType(LOGIN_SUCCESS).switchMap(() =>
+    socket$.map(message => {
+      switch (message.type) {
```
teddylo57: Nice, this seems cool

frontend/src/features/websocket/index.js
```diff
@@ -0,0 +1,61 @@
...
+            get(message, 'data.request.data.username')
+          );
+
+        case WS_SIT_DOWN_SUCCESS:
+          return convertWebsocketMessage(
+            SIT_DOWN_SUCCESS,
+            get(message, 'data.request.data.seatNumber')
+          );
+
+        default:
```
teddylo57: On the `backend`, it defaults to if the `messageType` isnt recognized then it throws an Exception - not sure if there is some refactoring needed there

ericliu121187: Right, this is switch statement is us receiving messages from the websocket, not sending -- so here we handle any messages that we have received that have types that do not match to the above two

PR #50, opened 2018-03-25
Add stand-up functionality on the BE for request and response to the FE

teddylo57:
-Added listener for `WS_STAND_UP_KICKOFF` and will perform `Table` actions such as decrement `playersize` and clear the occupied `seat`
-Added tests to for stand up to contrast sit down

One thing I added as a `TODO` is that we should check for the `state` of the `Table` before letting a player `stand`, whether in a hand or in a tournament, the functionality will be locked. But I also suppose, the FE would not display the `stand` button if in either of those states. So maybe an non-issue. Good to have backup code in the BE just in case though.

I do check if the seat is `occupied` before letting a player stand. I do this for syncing purposes. In case they sit/stand so fast that the `kickoff`, `success` messages are out of sync. Thoughts?


backend/src/main/java/poker/models/game/Table.java
```diff
@@ -35,6 +35,12 @@ public void sitDown(int seatNumber, Player player) {
     playerSize++;
   }
 
+  public void standUp(int seatNumber, Player player) {
```
ericliu121187: Yeah, I agree here we should check to see if it is the expected user that is sitting here

backend/src/main/java/poker/services/users/UserService.java
```diff
@@ -103,6 +103,30 @@ public static void sitUser(Session user, Request request) throws IOException {
...
+  public static void standUser(Session user, Request request) throws IOException {
+    Gson gson = new Gson();
+    int seatNumber = request.getData().getSeatNumber();
+
+    Response response = new Response();
+    Response.Data sitResponseData = response.new Data();
+
+    // TODO: add logic to check for tournament start here, dont allow player to stand if a tournament or hand is in session
+    if (!Api.testTable.getSeats().stream().filter(seat -> seat.getSeatNumber() == seatNumber)
+        .findFirst().get().isAvailable()) {
```
ericliu121187: I don't think we need this, I think we can get away with just returning success if the user stood up even though they aren't sitting

backend/src/main/java/poker/models/game/Table.java
teddylo57: added

backend/src/main/java/poker/models/game/Table.java
```diff
@@ -35,6 +35,14 @@ public void sitDown(int seatNumber, Player player) {
...
+    Seat seat = seats.get(seatNumber - 1);
+    if (seat.getPlayer().equals(player)) {
```
ericliu121187: Will this assume that the player we pass from the FE has exactly the same properties and values as the one currently on the BE? 

Will it be better to check if the username is equal?

teddylo57: Oops, I was drunk

PR #53, opened 2018-03-26
Resync table

teddylo57:
This PR sends messages to the FE after every `table interaction` and will resync the FE with the current state of the `table` on the backend. 

The messageType to be expected here is `WS_UPDATE_TABLE`.


teddylo57:
https://drive.google.com/open?id=1ah6e1RvwIPS5qVztTVs8DYm4z3VtAMjK

teddylo57:
Ah, do you want the delta instead?

PR #58, opened 2019-01-15
Sit down

ericliu121187:
https://trello.com/c/g9imjK4H/223-3-user-can-sit-down


frontend/src/features/table/index.js
```diff
@@ -0,0 +1,103 @@
...
+
+// EPICS
+
+const sitDownKickoffEpic = action$ =>
+  action$
+    .ofType(SIT_DOWN_KICKOFF)
+    .take(1)
+    .mergeMap(({ payload }) => {
+      return Observable.of(
+        sendWebsocketMessage(WS_SIT_DOWN_KICKOFF, { seat: payload.seat })
```
ericliu121187: @teddylo57 We're sending this up as `messageType`. We would be looking for the corresponding message type of `WS_SIT_DOWN_SUCCESS` or `WS_SIT_DOWN_FAILURE` back.

PR #74, opened 2019-02-05
hole cards + actions

teddylo57:
-add pot object to be tracked during betting
-add bet tracking per player for the round & hand
-removed player object from being sent repetitively, instead opting to use the username
-add support for hole cards being different per user, however if you are standing(spectating) you can see everyones hole cards
-add the framework for handState


backend/src/main/java/poker/services/gameplay/GameService.java
```diff
@@ -25,50 +29,94 @@ public static void startTournament() throws IOException {
...
     if (Api.testTable.getSeats().stream().map(Seat::getPlayer).filter(Objects::nonNull).count()
         >= 2) {
-      Api.testTable.start();
-      UserService.sendAll(String.valueOf(new JSONObject()
-          .put("serverMessage", "[SERVER] " + " Game is starting")));
+      Api.testTable.startGame();
+      UserService.sendAll(
+          String.valueOf(new JSONObject().put("serverMessage", "[SERVER] " + " Game is starting")));
       sendUpdatedTable();
+      sendAction();
```
ericliu121187: Could this occur in `sendUpdatedTable`, no need for a separate message?

PR #80, opened 2019-02-09
Can check

teddylo57:
-FE sends check, BE processes check option
-BE sends update table


ericliu121187:
Did you have a chance to test this with the FE?

teddylo57:
changed the field to be `type` so it works now

PR #83, opened 2019-02-09
Bet

teddylo57:
-add functionality to receive a bet


backend/src/main/java/poker/enums/game/Action.java
```diff
@@ -3,5 +3,8 @@
 public enum Action {
   FOLD,
   CHECK,
-  BET;
+  BET,
```
ericliu121187: Sooner or later, we'll want these two files to be synced up

https://github.com/ericliu121187/short-deck-poker/blob/master/frontend/src/constants/playerActionTypeEnum.js#L6

backend/src/main/java/poker/models/game/Player.java
```diff
@@ -50,4 +52,12 @@ public void check() {
     this.toAct = false;
   }
 
+  public void bet(int betAmount) {
+    this.toAct = false;
+    Api.testTable.pot.addToPot(UserService.getPlayerFromUsername(username), betAmount);
+    Api.testTable.setCurrentRoundBetAmount(betAmount);
```
ericliu121187: In the future, we may want to move these to hand instead of table.

teddylo57: Sure, can you create a ticket for it?

backend/src/main/java/poker/enums/game/Action.java
teddylo57: Synced

backend/src/main/java/poker/models/game/Player.java
ericliu121187: https://trello.com/c/clDYqSLQ/247-move-certain-fields-from-table-into-hand-model

PR #94, opened 2019-02-14
Show cards in showdown

teddylo57:
-if users are in showdown, show cards for 8 seconds
-send a list of winners to the FE once a hand has ended -> List<String> winners


backend/src/main/java/poker/models/game/Hand.java
```diff
@@ -45,6 +47,7 @@ public void dealNewHand(Table table) {
   }
 
   public void cleanHandData() {
+    showdown = false;
```
ericliu121187: sharing with you preference -- if it's boolean, i like to start it with a variable like `isShowingDown` or `couldBet` or `inShowdown`

teddylo57: changed

PR #99, opened 2019-02-15
fix bet amounts, fix fe red dot, make raise larger than bet

teddylo57:
-fix correct bet values
-fix FE red dot 
-fix FE to hardcode raise amount larger than bet (150 raise, 100 bet) for testing
-fix issue when flop was called twice


frontend/src/components/room/actions.js
```diff
@@ -94,7 +94,7 @@ function Actions({
             return (
               <ButtonWrapper key={possiblePlayerAction}>
                 <Button
-                  onClick={() => playerAction(playerActionTypeEnum.RAISE, 100)}
+                  onClick={() => playerAction(playerActionTypeEnum.RAISE, 150)}
```
ericliu121187: nice

frontend/src/components/room/table/seat.js
```diff
@@ -87,7 +87,6 @@ const DealerButton = styled.div`
 `;
 
 const DealerButtonWrapper = styled.div`
-  border: 1px solid red;
```
ericliu121187: ty

PR #101, opened 2019-02-15
add ability to fast forward to show down when users are put all in

teddylo57:
-fast forward to showdown if only one user is not all in with chips
-fix more betting


backend/src/main/java/poker/models/game/Table.java
```diff
@@ -116,6 +116,18 @@ public void setButton(int seatNumber) {
...
   }
 
+  List<Seat> getSeatsInHandWithChips() {
+    return getOccupiedSeats().stream()
+        .filter(seat -> seat.getPlayer().isInHand())
+        .filter(seat -> seat.getPlayer().getChips() > 0)
+        .collect(Collectors.toList());
+  }
+
+  boolean fastFowardToShowdown() {
```
ericliu121187: shouldFastFowardToShowdown

PR #105, opened 2019-02-24
User sees full-screen background

ericliu121187:
This provides a more polished look for the user.

screen shot 2019-02-23 at 7 53 46 pm


frontend/src/features/session/index.js
```diff
@@ -155,7 +155,7 @@ const initialState = {
   error: '',
   isLoggedIn: false,
   isLoggingIn: false,
-  username: 'Player123',
+  username: '',
```
ericliu121187: Removes `Player123` as the default username

teddylo57: yay

PR #108, opened 2019-02-24
User can see server messages

ericliu121187:
screen shot 2019-02-24 at 1 05 27 am


frontend/src/flow/types.js
```diff
@@ -24,6 +24,10 @@ declare type Card = {
   suit: Suit,
 };
 
+declare type ChatMessage = {
+  text: string,
```
ericliu121187: Please structure the JSON in the same convention as other ones, with the following shape for the data.

```json
data: {
  message: {
    text: string,
  }
}
```

frontend/src/features/chat/index.js
```diff
@@ -0,0 +1,41 @@
+/* @flow */
+
+import { createSelector } from 'reselect';
+import messageTypeEnum from 'constants/messageTypeEnum';
+
+const { WS_CHAT_MESSAGE } = messageTypeEnum;
```
ericliu121187: This is the suggested message type.

frontend/src/flow/types.js
teddylo57: How do we differentiate between `serverMessages`and `userMessages`?

PR #109, opened 2019-02-24
fix betting system

teddylo57:


backend/src/main/java/poker/models/game/Player.java
```diff
@@ -134,8 +134,8 @@ public void callAllIn() {
   }
 
   private void addAmount(int callAmount) {
-    if (callAmount > chips) {
-      Api.testTable.getHand().setRoundAmountToPlay(callAmount);
+    if (callAmount < chips) {
```
ericliu121187: A. Should `callAmount` be renamed to `amount`?
B. This if-then seems unnecessary?

PR #118, opened 2019-02-25
User has larger buttons, can see cards

ericliu121187:
screen shot 2019-02-24 at 10 33 20 pm


frontend/src/components/room/table/seat.js
```diff
@@ -184,7 +184,7 @@ function Seat({
 }) {
   const { roundBetAmount } = player || {};
   const isSeatable = !player;
-  const label = isSeated ? `Empty Seat` : `Seat ${number}`;
+  const label = isSeated ? `Empty Seat` : `Seat`;
```
ericliu121187: Removing seat number bc there are only 2 seats

teddylo57:
I like this a lot

PR #128, opened 2019-03-02
fix bad table state

teddylo57:


backend/src/main/java/poker/Routes.java
```diff
@@ -25,7 +25,7 @@
   private void userRoutes() {
     post("/login", "application/json", (request, response) -> {
       response.type("application/json");
-      Player player = gson.fromJson(request.body(), Player.class);
+      Player player = gson.fromJson(request.body().toLowerCase(), Player.class);
```
ericliu121187: this is gonna fix it? it's as simple as a unique username issue? i usually always just input all lowercase

teddylo57: think there may be another issue but this is a good check to have

PR #131, opened 2019-03-02
User is unable to make multiple player actions within 1s

ericliu121187:
We throttle the player action and take the first action.


frontend/src/features/player-action/index.js
```diff
@@ -98,10 +98,13 @@ export function deriveValidRaiseToAmount({
...
 
 const playerActionEpic = action$ =>
-  action$.ofType(PLAYER_ACTION).mergeMap(({ payload }) => {
-    const { amount, type } = payload;
-    sendWebsocketMessage(WS_PLAYER_ACTION, { amount, type });
-    return Observable.empty();
-  });
+  action$
+    .ofType(PLAYER_ACTION)
+    .throttleTime(1000, undefined, { leading: true })
```
ericliu121187: This is the only change -- the rest is just auto-formatted by prettier

PR #134, opened 2019-03-04
Refund bets, fix blinds timer, allow ff to showdown if blinded all-in

teddylo57:
-refund bets after showdown
-fix blinds timer
-allow fast forward to showdown if you are blinded all-in


backend/src/main/java/poker/models/game/Table.java
```diff
@@ -140,6 +141,28 @@ public void setButton(int seatNumber) {
         .collect(Collectors.toList());
   }
 
+  List<Seat> getSeatsRemainingToActAndAllIn() {
```
ericliu121187: nice, descriptive name

backend/src/main/java/poker/models/game/Table.java
```diff
@@ -140,6 +141,28 @@ public void setButton(int seatNumber) {
...
+    return getSeatsInHand().stream()
+        .filter(seat -> seat.getPlayer().isYetToAct())
+        .filter(seat -> seat.getPlayer().isAllIn())
+        .collect(Collectors.toList());
+  }
+
+  public boolean allInFastForwardBlinds() {
```
ericliu121187: something like `doBlindsPutPlayersAllIn`? not entirely sure what `allInFastForwardBlinds` means

backend/src/main/java/poker/models/game/Table.java
```diff
@@ -140,6 +141,28 @@ public void setButton(int seatNumber) {
...
+    if (getSeatsRemainingToActAndAllIn().size() > 0) {
```
ericliu121187: would another way to approach this issue to set player `isAllIn` to `true` when setting the blinds?

then we could re-use `shouldFastForwardToShowdown`?

backend/src/main/java/poker/models/game/Table.java
```diff
@@ -247,6 +271,18 @@ public void evaluateWinner(List<Player> players) {
     players.stream()
         .map(Player::getUsername)
         .forEach(s -> Logger.getLogger(Table.class).info("[WINNERS] " + s));
+
+    List<Player> sortedPlayers =
+    getSeatsInHand().stream().map(Seat::getPlayer).sorted(new TotalBetAmountComparator()).collect(Collectors.toList());
+
+    if (sortedPlayers.size() >= 2 && (sortedPlayers.get(0).getTotalHandBetAmount()) > sortedPlayers.get(1).getTotalHandBetAmount()) {
```
ericliu121187: 1 step closer to side pots

backend/src/main/java/poker/models/game/Table.java
```diff
@@ -298,6 +334,7 @@ public void reset() {
               player.setYetToAct(false);
               player.setAllIn(false);
             });
+    this.increaseBlinds = false;
```
ericliu121187: how come we are removing this?

backend/src/main/java/poker/models/game/Table.java
teddylo57: Hm, yeah but if the player that is all-in is the big blind, we still want the other player to act, and to call the big blind.

If the player that is all-in is the small blind, we want to fast forward

backend/src/main/java/poker/models/game/Table.java
teddylo57: Adding it in, to reset the table to the initial state

backend/src/main/java/poker/models/game/Table.java
ericliu121187: got it -- from a product perspective, i can't think of a situation where a player who is all-in can still act

teddylo57: i'll add that in the refactor, cleaning up the code

PR #143, opened 2019-03-05
add changelog to track backend changes

teddylo57:


teddylo57:
Just to keep a record of what goes in each build, currently we're in development so it's not super important because we haven't released but in the future it'll help distinguish builds when we release less frequently

PR #150, opened 2019-03-08
Users can only input numbers and increment/decrement by big blind amount for bet sizing

ericliu121187:
This is the improvement I should have went with from the beginning. This is a major value add to the user as it improves two flows -- 

A. If the user uses only keyboard -- user can use up/down buttons and then hit enter.
B. If the user uses only the mouse -- user can click up/down buttons and then click the action button.

https://cl.ly/b36c1e55496e


frontend/src/components/room/actions/amount-field.js
```diff
@@ -33,7 +33,16 @@ class AmountField extends Component<void, void, void> {
...
 
   render() {
-    return <AmountFieldInput ref={this.inputRef} {...this.props} />;
+    const { bigBlindAmount } = this.props;
+
+    return (
+      <AmountFieldInput
+        ref={this.inputRef}
+        step={bigBlindAmount}
+        type="number"
```
ericliu121187: this was very nice -- minimal research unearthed a super low LOE solution -- in HTML5

type and step attributes ✅

PR #161, opened 2020-04-16
Sessions management with UUID

teddylo57:
-Renamed the FE fields to uuid to keep in line with UUID name

-BE now keeps a map of current sessions in the form of Map<Sessions, UUID> and also has a map of all users in the form of Map<UUID, Player> which in the future we can use a LRU cache to keep track of that.

Note: Users will keep their chip amount because that is stored by UUID, they need not to login with the same username. 
However, I saw a bug that I'll track with a trello ticket when users change their username


backend/src/main/java/poker/Api.java
```diff
@@ -1,14 +1,15 @@
...
-import java.util.Properties;
 import poker.enums.hand.BlindLevel;
 import poker.models.game.Table;
 import poker.services.CorsFilter;
 import poker.services.users.UsersWebsocketHandler;
 
+import java.io.IOException;
+import java.util.Properties;
+
+import static spark.Spark.webSocket;
```
ericliu121187, line 11: Are you using a linter that auto-sorts these imports? I would need to install it

backend/src/main/java/poker/Routes.java
```diff
@@ -25,20 +24,15 @@
   private void userRoutes() {
     post("/login", "application/json", (request, response) -> {
       response.type("application/json");
-      Player player = gson.fromJson(request.body().toLowerCase(), Player.class);
-      LOGGER.info("Player: " + player.getUsername() + " is trying to login");
-      if (!UserService.isValidUsername(player.getUsername())) {
-        LOGGER.info(player.getUsername() + " is not a valid username");
+      Player connectedPlayer = gson.fromJson(request.body().toLowerCase(), Player.class);
```
ericliu121187, line 27: This player wouldn't be logged in or connected at this point, right?

backend/src/main/java/poker/models/game/Player.java
```diff
@@ -1,23 +1,22 @@
...
+import java.io.Serializable;
+import java.util.*;
+
 @Data
 public class Player extends Observable implements Serializable {
 
   // TODO: update starting_chips
   final static public int STARTING_CHIPS = 5000;
   private static final Logger LOGGER = Logger.getLogger(Player.class);
+  private UUID uid;
```
ericliu121187, line 19: Cool, there's a type for this!

backend/src/main/java/poker/services/users/UserService.java
```diff
@@ -19,12 +13,21 @@
...
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.stream.Collectors;
+
 public class UserService {
 
   private static final Gson gson = new Gson();
   private static final Logger LOGGER = Logger.getLogger(UsersWebsocketHandler.class);
   // this map is shared between sessions and users, so it needs to be thread-safe (http://stackoverflow.com/a/2688817)
-  static Map<Session, Player> userUsernameMap = new ConcurrentHashMap<>();
+  static Map<Session, UUID> sessionUUIDMap = new ConcurrentHashMap<>();
```
ericliu121187, line 29: I do see that in the past it's Session to Player. How come it's not a UUID to Session?

backend/src/main/java/poker/services/users/UserService.java
```diff
@@ -69,20 +72,38 @@ static void connectUser(Session user, Request request) {
...
+    LOGGER.info("[WEBSOCKET] UUID " + connectedUUID + " username: " + connectedUsername + " is connecting");
+    System.out.println("UUID: " + connectedUUID);
+
+    if (playersMap.containsKey(connectedUUID)) {
+      playersMap.get(connectedUUID).setUsername(connectedUsername);
+      if (Api.testTable.getOccupiedSeats().stream()
+              .anyMatch(seat -> seat.getPlayer().getUid().equals(connectedUUID))) {
+        LOGGER.info("[RECONNECT] " + connectedUsername + " has reconnected");
+
+        UserService.sendServerMessage(UserService.getPlayer(user).getUsername()
```
ericliu121187, line 87: 👌What may be interesting (this warrants a larger conversation to see if we should even allow players to change usernames) is "Steve has reconnected as Eric." so we show what the previous username was

backend/src/main/java/poker/services/users/UserService.java
```diff
@@ -150,25 +171,18 @@ public static boolean isValidUsername(String username) {
         !Character.isDigit(username.charAt(0));
   }
 
-  public static boolean isAvailable(String username) {
```
ericliu121187, line 153: nit: because this is UserService (as opposed to UsernameService), naming this "isUsernameAvailable" is a bit more descriptive

backend/src/main/java/poker/services/users/UserService.java
```diff
@@ -150,25 +171,18 @@ public static boolean isValidUsername(String username) {
...
-    return userUsernameMap.entrySet().stream().filter(key -> key.getValue().equals(player))
+  public static Session getSession(UUID uuid) {
+    return sessionUUIDMap.entrySet().stream().filter(key -> key.getValue().equals(uuid))
         .findFirst().get().getKey();
   }
 
   public static List<Player> getConnectedPlayers() {
-    return new ArrayList<>(userUsernameMap.values());
+    return sessionUUIDMap.values().stream().map(s -> playersMap.get(s)).collect(Collectors.toList());
+//    return new ArrayList<>(userUsernameMap.values());
```
ericliu121187, line 185: Remove?

backend/src/main/java/poker/services/users/UsersWebsocketHandler.java
```diff
@@ -25,50 +23,24 @@
   public void onConnect(Session user) {
     LOGGER.info("[WEBSOCKET] Connection received");
     LOGGER.info("[OnConnect] Received");
-    if (Api.testTable.getOccupiedSeats().stream()
```
ericliu121187, line 28: How come we're moving this logic to the UserService?

backend/src/main/java/poker/Api.java
teddylo57: I just use organize imports built into intelliJ, but I use google java styles for formatting and just auto-formats all the files => https://google.github.io/styleguide/javaguide.html

backend/src/main/java/poker/Routes.java
teddylo57: No not yet, I just get the username from the data object, but maybe I don't need to cast it into a Player type or I should rename the variable

backend/src/main/java/poker/models/game/Player.java
teddylo57: Yeah I wanted to use UUID because it's built-in to Java

ericliu121187: Could you have two UUID types on a class? for example, for `uid` and `socialSecurityNumber`

backend/src/main/java/poker/services/users/UserService.java
teddylo57: I think we could map UUID to Session, what do you think the advantages would be?

backend/src/main/java/poker/services/users/UserService.java
teddylo57: Sure, I can add a change for that

backend/src/main/java/poker/services/users/UserService.java
teddylo57: Will do

backend/src/main/java/poker/services/users/UserService.java
ericliu121187: Not sure how maps work in Java - if it is `<UUID, Session>`, would UUID be the index? If so, that would be a faster look up if we're trying to see if there is a session already for the `uid`?

backend/src/main/java/poker/models/game/Player.java
teddylo57: Yeah, I don't see why we couldn't, I don't think UUID acts as a primary key for any field

backend/src/main/java/poker/services/users/UserService.java
teddylo57: Hm, I just tested it out and I was under the impression that a different tab would have a different uuid so I thought session would be more unique however, it seems uuid is the same across tabs. I can make this change

ericliu121187: Ah, yes! Cookies are domain-based so they will be the same across domains. We _could_ have diff sessions in the same domain (then we would need to make it domain + path) - right now I have a hard time thinking of what we would need to do that.

Ofc, since cookies are stored in the browser, for dev purposes or new session purposes, we can open an incognito window which wouldn't have the cookie stored.

teddylo57: Cool, I'll implement the changes

backend/src/main/java/poker/services/users/UserService.java
teddylo57: I think username available suggest that it will check if anyone else has the username, but this function actually checks if the formatting is correct and between 3-20 characters long

PR #163, opened 2020-04-22
Chat support 

teddylo57:
-Added two message types: `WS_SEND_CHAT_MESSAGE_KICKOFF` and `WS_SEND_CHAT_MESSAGE_SUCCESS`
-Allows sending of messages from the FE
-Removed `Dealer:` prefix from FE and made different functions on the BE to differentiate regular chat messages vs. server/dealer messages


frontend/src/components/room/chat/index.tsx
```diff
@@ -77,7 +77,7 @@ function Chat({ messages }: { messages: Record<string, any>[] }) {
           return (
             // eslint-disable-next-line react/no-array-index-key
             <MessageWrapper key={`${text}-${i}`}>
-              <Message>{`Dealer: ${text}`}</Message>
```
ericliu121187, line 80: Nice catch!

PR #165, opened 2020-04-24
Message Queue for users that reconnect

teddylo57:
-Users that reconnect will receive the last 20 chat messages sent (dealer/player)


backend/src/main/java/poker/services/users/UserService.java
```diff
@@ -99,6 +101,12 @@ static void messageSuccess(Request request) {
     UserService.sendMessageToUser(UUIDSessionMap.get(sender), gson.toJson(response));
   }
 
+  static void sendMessageQueue(UUID uuid) {
+    for (Response response : responseQueue.getQueue()) {
+      UserService.sendMessageToUser(UUIDSessionMap.get(uuid), gson.toJson(response));
```
ericliu121187, line 106: Nice, so 20 messages (max) sent to frontend upon reconnect

teddylo57: Yup, we can change the number in the future if needed

PR #167, opened 2020-04-26
User tries to automatically reconnect upon disconnect

ericliu121187:
Release Notes

This will start our switch over to a more robust websocket connection. We actually always have an open ws connection. Upon login, we could change our `messageType` instead of `WS_CONNECT_KICKOFF` to `WS_LOGIN_KICKOFF`.

I tested this by simply not having the API running.

Note that I was not able to use reconnecting-websocket, because it does not seem to integrate easily with RxJS.

Screenshots

https://share.getcloudapp.com/8LujPL88


frontend/yarn.lock
```diff
@@ -1761,11 +1761,6 @@ abab@^2.0.0:
   resolved "https://registry.yarnpkg.com/abab/-/abab-2.0.3.tgz#623e2075e02eb2d3f2475e49f99c91846467907a"
   integrity sha512-tsFzPpcttalNjFBCFMqsKYQcWxxen1pgJR56by//QwvJc4/OUS3kPOOttx2tSIfjsylB0pYu7f5D3K1RCxUnUg==
 
-abbrev@1:
```
ericliu121187, line 1764: These are minor updates that happened because I tried out `reconnecting-websocket`.

PR #169, opened 2020-04-28
Add heartbeat response `pong` 

teddylo57:
-Added `WS_PONG` type into a response message for the FE
-I've commented out the messageQueue for the time being until we can debug the /connect issue to prevent overflow of the queue
-Refactored some methods to send by UUID param
-Null check for request type
Screen Shot 2020-04-28 at 5 57 47 PM


backend/src/main/java/poker/services/users/UserService.java
```diff
@@ -98,15 +103,26 @@ static void messageSuccess(Request request) {
...
+    Response response = new Response();
+    Response.Data messageResponseData = response.new Data();
+    messageResponseData.setSuccess(true);
+    messageResponseData.setRequest(request);
+    response.setType(MessageType.WS_PONG);
+    response.setData(messageResponseData);
+    UserService.sendMessageToUser(sender, gson.toJson(response));
   }
 
+//  static void sendMessageQueue(UUID uuid) {
```
teddylo57, line 120: Commented out for the time being

backend/src/main/java/poker/services/users/UserService.java
```diff
@@ -120,7 +136,7 @@ static void connectUser(Session user, Request request) {
     System.out.println("UUID: " + connectedUUID);
 
     if (playersMap.containsKey(connectedUUID)) {
-      sendMessageQueue(connectedUUID);
+//      sendMessageQueue(connectedUUID);
```
teddylo57, line 139: Temporarily commented out

backend/src/main/java/poker/services/users/UserService.java
```diff
@@ -31,27 +34,29 @@
   static Map<UUID, Player> playersMap = new ConcurrentHashMap<>();
   static MessageQueue<Response> responseQueue = new MessageQueue<>(QUEUE_MAX_SIZE);
 
-  public static void sendMessageToUser(Session user, String message) {
-    LOGGER.info("Sending to: " + getPlayer(user).getUsername());
+  public static synchronized void sendMessageToUser(UUID uuid, String message) {
```
ericliu121187, line 37: Not familiar with `synchronized`. Does it lock this method? Why do we need it?

backend/src/main/java/poker/services/users/UsersWebsocketHandler.java
```diff
@@ -58,44 +58,51 @@ public void onMessage(Session user, String message) throws Exception {
...
-      case WS_PLAYER_ACTION:
-        GameService.processPlayerAction(user, request);
-        break;
-      case WS_RESET_TABLE:
-        GameService.resetTournament();
-        break;
-      default:
-        LOGGER.error(request.toString());
-        LOGGER.error("MessageType was not recognized");
+    if (request.getType() == null) {
```
ericliu121187, line 61: I don't think the message the frontend was sending had a `null` type because it had `WS_PONG`. Were you able to repro the server closing the websocket connection?

backend/src/main/java/poker/services/users/UserService.java
teddylo57: Hm, it was a fix before I uncommented the message queue, due to the rapid reconnect/disconnect every few seconds, the queue had to send 20+ messages all the time. 

So right now, we don't need it, but in the future we might want to so it locks methods to threads

backend/src/main/java/poker/services/users/UsersWebsocketHandler.java
teddylo57: Yup, before it would try to get `request.getType()` but the type was not in the enum, therefore it was null

backend/src/main/java/poker/services/users/UserService.java
ericliu121187: Ya, would it slow us down now, given that this method is locked and we need to send to multiple ppl?

backend/src/main/java/poker/services/users/UsersWebsocketHandler.java
ericliu121187: I see if it's not part of the type it would change it to null huh, interesting

teddylo57: https://stackoverflow.com/questions/23056324/why-does-java-allow-null-value-to-be-assigned-to-an-enum

backend/src/main/java/poker/services/users/UserService.java
teddylo57: Yeah we can tinker with this method but during my testing it worked ok

backend/src/main/java/poker/services/users/UsersWebsocketHandler.java
ericliu121187: Okay for my own clarification - if we try to `getType()` and the type field is `"FOOBAR"`, because `"FOOBAR"` isn't in the enum, `getType()` would return `null`?

teddylo57: Yes, we can either create a UNKNOWN variable to prime the enum value, and have it overwritten by request.getData() or we can have the null check

PR #170, opened 2020-04-29
Move request object up a level from data 

teddylo57:
per https://trello.com/c/7AzysZmj/362-bug-fix-connect-is-being-sent-too-often

-Changes the model of a response from the BE to 
```{
  data: // what the message returns
  request: // [optional] a copy and paste of what the request was if any
  type: // message type
}


ericliu121187:
After this is merged, `request` has been moved up a level ✅but I'm not seeing the expected request model of the frontend request. 

https://share.getcloudapp.com/7Ku0PQkR

PR #171, opened 2020-04-30
Frontend sends `cid` for `WS_PING`, handles WS_CONNECT_KICKOFF/WS_CONNECT_FAILURE

ericliu121187:
@teddylo57 I did not do much testing w this, please lmk if there are any bugs the backend sees


frontend/src/features/session/index.js
```diff
@@ -199,6 +199,7 @@ export default function sessionReducer(
       };
 
     case LOGIN_FAILURE:
+    case WS_CONNECT_FAILURE:
```
ericliu121187, line 202: Please note that this is the message type if there is a failure that frontend expects from backend

PR #172, opened 2020-04-30
Add cid

teddylo57:
-Deserialized cid field
-Need to know format for expected Heartbeat response, cid, and uid?


ericliu121187:
For heartbeat, could we get away with just `cid` for now? We can add `uid` later if needed

backend/src/main/java/poker/models/websocket/HeartbeatResposne.java
```diff
@@ -0,0 +1,15 @@
+package poker.models.websocket;
+
+import lombok.Data;
+import poker.enums.websocket.MessageType;
+
+@Data
+public class HeartbeatResposne {
```
ericliu121187, line 7: `HeartbeatResponse`?

frontend/src/features/session/index.js
```diff
@@ -12,6 +12,7 @@ import SessionCookiesEnum from './session-cookies-enum';
 import { v4 as uuidv4 } from 'uuid';
 
 const {
+  WS_CONNECT_FAILURE,
```
ericliu121187, line 15: 🙏

backend/src/main/java/poker/models/websocket/HeartbeatResposne.java
teddylo57: Yeah, I think to remove all those other fields, I'll create a specific class for heartbeat response with the fields data, cid, type

teddylo57:
How's `{"data":{"success":true},"cid":"2020-04-30T21:09:58.432Z","type":"WS_PONG"}`

ericliu121187:
Yeah I think that works - question - I forgot why we had `"success": true` in responses, when we use `_FAILURE` or is that when the data is empty we include it? Or just in this case?

teddylo57:
Yeah, I don't think there is a need for both - it's whatever the FE rather parse out, either success or _FAILURE/_SUCCESS

backend/src/main/java/poker/models/websocket/HeartbeatResposne.java
teddylo57: Let me know if it's okay to remove {data}, as success is not needed right?

ericliu121187: Prob okay ya

PR #176, opened 2020-05-07
Update user's handrank

teddylo57:
Handrank will be sent to an user with the format 
```
{"data":{"handRank":"STRAIGHT"},"type":"WS_HAND_RANKING"}

### CODE REVIEW

File: backend/src/main/java/poker/services/gameplay/HandRankService.java [hunk 1]
```diff
@@ -0,0 +1,96 @@
...
+  @Override
+  public void update(Observable o, Object arg) {
+    updateEvaluator(table.getHand().getHandState());
+    updateHandRanks();
+  }
+
+  private void updateEvaluator(HandState state) {
+    for (Seat seatedPlayer : table.getSeatsInHand()) {
+      switch (state) {
+        case NOT_STARTED:
```
ericliu121187, line 30: 1. Should we add `PRE_FLOP` to this?
1. Does Java support this:

```java
case NOT_STARTED:
case STARTING:
case PRE_FLOP:
case END:
  break;
```

backend/src/main/java/poker/models/websocket/Response.java
```diff
@@ -2,6 +2,7 @@
 
 import lombok.Data;
 import poker.enums.game.Action;
+import poker.enums.hand.Handrank;
```
ericliu121187, line 5: To confirm all the enum values here are all the possible values of `handRank`? Are there any values outside of this?

backend/src/main/java/poker/services/gameplay/HandRankService.java
teddylo57: 1. Wanted Pre-flop, Flop, turn, river to be caught in the default to update hand rank.
2. Yes, I should have done the second one

ericliu121187: What value(s) would `PRE_FLOP` show given that it is only 2 cards?

backend/src/main/java/poker/models/websocket/Response.java
teddylo57: Yup,   ROYAL_FLUSH(10),
  STRAIGHT_FLUSH(9),
  FOUR_OF_A_KIND(8),
  FLUSH(7),
  FULL_HOUSE(6),
  STRAIGHT(5),
  THREE_OF_A_KIND(4),
  TWO_PAIR(3),
  ONE_PAIR(2),
  HIGH_CARD(1);

ericliu121187: Sounds good! Let's start w this

backend/src/main/java/poker/services/gameplay/HandRankService.java
teddylo57: High Card or One Pair

ericliu121187: Sounds good!

PR #178, opened 2020-05-07
Frontend sends `uid` on WS_CONNECT_KICKOFF, extend heartbeat interval to 30000 ms

ericliu121187:
In order for us to bring the user back to their previous state, we need to send `uid` on `WS_CONNECT_KICKOFF` and pull the previous username.

This PR now sends it over (https://share.getcloudapp.com/DOuQd2Db).

What we need to do:

When we send `uid`, the backend needs to send back `WS_CONNECT_SUCCESS` (when there is a valid username) or `WS_CONNECT_FAILURE` (when there is an invalid username)

Here are the response scenarios
1. If there is a username sent and it matches up to the uid in the backend, return `WS_CONNECT_SUCCESS` with that username.
1. If there is a username sent and it does not match up to the one in the backend, check to see if it is a valid one and not being taken by any other player. If so, return `WS_CONNECT_SUCCESS` with that username.
1. If there is no username sent and there is none in the backend, return `WS_CONNECT_FAILURE`.
1. If there is a username sent and it is invalid or taken by another player, return `WS_CONNECT_FAILURE`.


frontend/src/features/websocket/index.ts
```diff
@@ -139,7 +139,10 @@ const heartbeatEpic = (action$: any) =>
   action$
     .ofType(WebsocketActionTypes.CONNECT_TO_WEBSOCKET_SUCCESS)
     .switchMap(() => {
-      return Observable.timer(0, 3000).switchMap(() => {
+      // Using this to prevent a websocket disconnection due to a timeout
+      const HEARTBEAT_INTERVAL = 30000;
```
ericliu121187, line 143: Extend heartbeat interval

teddylo57: Good idea, I think heartbeat should just be used to keep connection alive. We can implement other ways to index messages and keep track of state

PR #179, opened 2020-05-08
Update WS_CONNECT scenarios

teddylo57:
-Updated scenarios for WS_CONNECTION based on previous FE PR https://github.com/ericliu121187/short-deck-poker/pull/178


backend/src/main/java/poker/services/users/UserService.java
```diff
@@ -137,47 +136,62 @@ static void sendMessageQueue(UUID uuid) {
...
-
-        UserService.sendDealerMessage(playersMap.get(connectedUUID).getUsername()
-                + " rejoined the table.");
+  static void sendConnectionResponse(Request request, MessageType messageType) {
+    if (messageType != null) {
+      Response response = new Response();
+      Response.Data connectResponseData = response.new Data();
+      connectResponseData.setCreatedAtUtc(Instant.now().toString());
+      response.setRequest(request);
+      switch (messageType) {
```
ericliu121187, line 145: Does `response.setType(messageType)` work?

teddylo57: Sure we could, switch was in case we want to make it more general in the future

PR #181, opened 2020-05-11
User has hand rank

ericliu121187:
User has hand rank shown.

We need to still figure out what the hand rank would be when the player folds or when they do not have a hand rank. For now, it's a small label on the cards, more styling to come.

https://share.getcloudapp.com/4guyNwng


teddylo57:
Looks good!

teddylo57:
Shows for both?

ericliu121187:
Good catch - will fix

PR #182, opened 2020-05-11
User sends uid on /login

ericliu121187:
In the future, we could probably rip out /login endpoint. Until then, frontend now passes up the `uid`.

https://share.getcloudapp.com/12uy7ex9


teddylo57:
Agreed, we can rip it when we implement lobby

PR #184, opened 2020-05-11
Reset table

teddylo57:
-Stand players up during reset table


teddylo57:
Yeah, the other change was minor so I worked off this branch

PR #188, opened 2020-05-12
User has loading screen

ericliu121187:
Hard to see in the video - now there is loading screen. Before, it would flicker to the login page and then flicker to the table.

https://share.getcloudapp.com/WnuG6p5d


teddylo57:
This is a good feature, but it breaks incognito login.

Loading should transition to /login if WS_CONNECT_FAILURE is sent

PR #190, opened 2020-05-13
fixes include sitting down with no chips

teddylo57:
-Reset for all players, not just seated
-Cannot sit with no chips


backend/src/main/java/poker/Api.java
```diff
@@ -14,7 +14,7 @@
 
   private static final Properties properties = new Properties();
 
-  public static Table testTable = new Table("SDP 10 ETH tournament", 1, BlindLevel.BLIND_25_50);
+  public static Table testTable = new Table("SDP Tournament", 1, BlindLevel.BLIND_25_50);
```
ericliu121187, line 17: Why did we call it 10 ETH? Because of 10 ethereum?? Haha

backend/src/main/java/poker/models/game/Table.java
```diff
@@ -336,23 +336,25 @@ public void run() {
...
-              player.setYetToAct(false);
-              player.setAllIn(false);
-            });
     for (Seat seat : seats) {
       if (seat.getPlayer() != null) {
         standUp(seat.getSeatNumber(), seat.getPlayer());
       }
     }
+
+    UserService.getAllPlayers().stream()
```
ericliu121187, line 345: JIC players stand up and sit back down? Or relogin?

backend/src/main/java/poker/services/gameplay/GameService.java
```diff
@@ -252,6 +252,7 @@ private static HoleCards getPlayersHoleCards(String username) {
 
   // TODO: Debug feature
   public static void resetTournament() throws IOException {
+    UserService.sendDealerMessage("Table has been reset. Please take a seat");
```
ericliu121187, line 255: 👌

backend/src/main/java/poker/services/users/UserService.java
```diff
@@ -224,11 +224,11 @@ static void sitUser(Session user, Request request) {
         .filter(seat -> seat.getSeatNumber() == seatNumber)
         .findFirst()
         .get()
-        .isAvailable()) {
+        .isAvailable() && playersMap.get(request.getData().getUid()).getChips() > 0) {
```
ericliu121187, line 227: From a product standpoint, to make it simpler:

1. Player can only sit back to their previous seat.
1. Once a seat is taken by a player, only that player can sit it.
1. If a new game is started, all seats are reset.

backend/src/main/java/poker/services/users/UserService.java
```diff
@@ -224,11 +224,11 @@ static void sitUser(Session user, Request request) {
...
       Api.testTable.sitDown(seatNumber, getPlayer(user));
       response.setType(MessageType.WS_SIT_DOWN_SUCCESS);
     } else {
-      sitResponseData.setError("That seat is already taken.");
+      sitResponseData.setError("You are unable to sit there. Please ensure there is no one seated there and you have the correct chip amount");
```
ericliu121187, line 231: Nit: from client perspective, the error message would be clearer and more usable if it included what "correct" meant. In this case, maybe the better message could be "You are unable to sit there. That seat might be taken already or you have no chips left."

PR #196, opened 2020-05-17
Frontend sends `actionId` and uses it to hide the action buttons if action is already done

ericliu121187:
The frontend is utilizing `actionId`. The feature does not work - it seems that backend is not liking the comparison (details below).


backend/src/main/java/poker/models/game/Table.java
```diff
@@ -29,15 +29,15 @@
   private transient BlindLevel blindLevel;
   private transient boolean increaseBlinds;
   private transient HandRankService handRankService;
-  private UUID messageId;
+  private UUID actionId;
```
ericliu121187, line 32: Minor refactoring

backend/src/main/java/poker/services/gameplay/GameService.java
```diff
@@ -180,8 +180,8 @@ public static void processPlayerAction(Session user, Request request) {
           LOGGER.error("Not recognized player action");
       }
     } else {
-      LOGGER.error("Message Id mismatch: " + request.getData().getActionId() + " " +  Api.testTable.getMessageId());
-      LOGGER.error("Expected: " +  Api.testTable.getMessageId());
+      LOGGER.error("Message Id mismatch: " + request.getData().getActionId() + " " +  Api.testTable.getActionId());
```
ericliu121187, line 183: This is triggering, but it's not apparent to me why - could it be due to types?

```
[10:36:26 ERROR GameService:183] Message Id mismatch: 582ef849-0bda-4e0e-8d38-ac9c8e61e919 582ef849-0bda-4e0e-8d38-ac9c8e61e919
[10:36:26 ERROR GameService:184] Expected: 582ef849-0bda-4e0e-8d38-ac9c8e61e919
[10:36:26 ERROR GameService:185] Actual: 582ef849-0bda-4e0e-8d38-ac9c8e61e919
```

PR #199, opened 2020-05-18
Fix 

teddylo57:
Forgot to commit one line


teddylo57:
When you get a chance, can you review?

PR #203, opened 2020-05-22
Fix user should only see their own hand rank during showdown

ericliu121187:
https://share.getcloudapp.com/DOuQZBZ7


ericliu121187:
Also added a bunch of favicons

ericliu121187:
@teddylo57 please review again when you can

PR #205, opened 2020-05-31
Send NULL handrank when the table is reset

teddylo57:
When the table is reset, BE sends a `NULL` value as handrank and updates.

FE will need to parse out `NULL` and have it blank


backend/src/main/java/poker/enums/hand/Handrank.java
```diff
@@ -11,7 +11,8 @@
   THREE_OF_A_KIND(4),
   TWO_PAIR(3),
   ONE_PAIR(2),
-  HIGH_CARD(1);
+  HIGH_CARD(1),
+  NULL(0);
```
ericliu121187, line 15: Could this be NOT_AVAILABLE instead of NULL because it is more product logic semantic?

backend/src/main/java/poker/services/gameplay/HandRankService.java
```diff
@@ -52,6 +52,12 @@ public void updateHandRanks() {
     }
   }
 
+  public void resetHandRanks() {
```
ericliu121187, line 55: `sendHandRanksToPlayers` might be clearer

backend/src/main/java/poker/models/game/Player.java
```diff
@@ -40,6 +40,7 @@ public Player(UUID uid, String username) {
     this.inHand = false;
     this.toAct = false;
     this.yetToAct = false;
+    this.evaluator.clearHandRank();
```
teddylo57, line 43: Adding this line to initialize HandRank to be `NOT_AVAILABLE`, and can be updated as hole cards, flop, etc come out

backend/src/main/java/poker/services/gameplay/HandRankService.java
teddylo57: Agreed, let me rename some methods

backend/src/main/java/poker/services/gameplay/HandRankService.java
```diff
@@ -46,12 +46,25 @@ private void update(Seat seatedPlayer) {
     seatedPlayer.getPlayer().setEvaluator(new Evaluator(playersHand));
   }
 
-  public void updateHandRanks() {
+  public void updateSeatedHandRanks() {
```
ericliu121187, line 49: What is the difference between this and `sendHandRanksToAllPlayers`? Also, could the naming convention be the same and start with `sendHandRanks-`?

ericliu121187: Discussed - this might be more consistent if it was named `sendHandRanksToSeatedPlayers`?

PR #206, opened 2020-05-31
Add username errors to /login response

teddylo57:
A list of errors will be returned on /login if there are any issues in the format ```{"errors":{"errorList":["Username cannot contain DEALER.","Username cannot start with a digit."]}}```


backend/src/main/java/poker/models/websocket/ErrorResponse.java
```diff
@@ -2,14 +2,15 @@
...
+
 @Data
 public class ErrorResponse {
 
   private Errors errors;
 
   @Data
   public class Errors {
-
-    private String username;
```
teddylo57, line 13: Removed username field as I think it is no longer necessary, if you want we can add the username sent for a more complete message but I think it is redundant

backend/src/main/java/poker/Routes.java
```diff
@@ -29,21 +31,22 @@ private void userRoutes() {
...
           }
         });
   }
 
-  private String loginError(String error) {
+  private String loginError(List<String> errors) {
     ErrorResponse errorResponse = new ErrorResponse();
     ErrorResponse.Errors usernameError = errorResponse.new Errors();
-    usernameError.setUsername(error);
+    usernameError.setErrorList(errors);
```
ericliu121187, line 49: Could the response model be:

```json
{
  errors: {
     username: String[], 
  }
}
```

backend/src/main/java/poker/services/users/UserService.java
```diff
@@ -256,10 +257,21 @@ public static boolean isValid(UUID uuid, String username) {
         && !Character.isDigit(username.charAt(0))
         && playersMap.values().stream()
             .filter(player -> !player.getUid().equals(uuid))
-            .noneMatch(player -> player.getUsername().toLowerCase().equals(username.toLowerCase()))
-        && playersMap.values().stream()
+            .noneMatch(player -> player.getUsername().toLowerCase().equals(username.toLowerCase()));
+  }
+
+  public static List<String> getLoginErrors(UUID uuid, String username) {
```
ericliu121187, line 263: `validateLogin` might be a more descriptive, I've seen that as convention.

teddylo57: Fixed

teddylo57:
Example message: 

```{
  "errors": {
    "username": [
      "Username cannot contain DEALER.",
      "Username cannot start with a digit."
    ]
  }
}```

File: backend/src/main/java/poker/models/websocket/ErrorResponse.java [hunk 4]
```diff
@@ -2,14 +2,16 @@
...
 
-  private Errors errors;
-
-  @Data
-  public class Errors {
+  private Map<String, List<String>> errors;
 
-    private String username;
-  }
+//  @Data
```
ericliu121187, line 13: Is this meant to be removed?

PR #207, opened 2020-05-31
Winner list

teddylo57:
Winner field now has the fields of `uid` `chipsWon` `username` `delta`


backend/src/main/java/poker/models/game/Winner.java
```diff
@@ -0,0 +1,19 @@
...
+import lombok.Data;
+
+import java.util.UUID;
+
+@Data
+public class Winner {
+    private UUID uid;
+    private int chipsWon;
+    private String username;
+    private int delta;
```
ericliu121187, line 12: Discussed that this could be out of the scope of this PR, provided that this part is a P3. Would this be clearer if this was named `chipsDelta`?

backend/src/main/java/poker/models/game/HandState.java
```diff
@@ -154,7 +154,7 @@ public HandState checkBets(Table table, HandState nextState) {
       if (checkForWinner(table)) {
         table.adjustWinnerChips();
         table.getHand().setPlayersInHandYetToAct(table.getSeatsInHand());
-        SHOWDOWN.performTask(table);
+        checkForEndState(table);
```
ericliu121187, line 157: Thanks for fixing this! Did not catch this - we could have a checklist for important use cases.

backend/src/main/java/poker/services/gameplay/HandRankService.java
```diff
@@ -67,15 +68,17 @@ public static void evaluateHands(Table table) {
 
     UserService.sendDealerMessage("-- Showdown -- ");
 
-    Evaluator max = Collections.max(hands, new HandRanker());
+    // get best hand on the board
+    Evaluator bestHand = Collections.max(hands, new HandRanker());
```
ericliu121187, line 72: Thanks for refactoring this

PR #219, opened 2020-06-28
Fix incorrect blinds level for BLINDS_STATUS by sending initial BLINDS_STATUS message for first hand

ericliu121187:
https://trello.com/c/dJyw9ImR/415-bug-user-has-correct-blind-level-in-the-blindsstatus

This also fixes an issue where blinds were incorrectly starting at 50/100.

Tested
- Works upon first start + table reset (https://share.getcloudapp.com/xQuDq0nX)


backend/src/main/java/poker/models/game/Hand.java
```diff
@@ -44,8 +44,14 @@ public Hand(Table table, HandState state) {
   public void dealNewHand(Table table) {
     this.handId++;
     this.deck = new Deck();
+    LOGGER.info("-------------- ERIC dealNewHand HAS BEEN HIT ---------------");
+    LOGGER.info(table.getBlindLevel());
     dealHoleCardsTo(table.getOccupiedSeats());
+
+    // TODO Could the logic of sending out chat messages could be in Table, triggered
+    // within startBlindTimer, deprecating the need to have a increaseBlinds boolean?
     if (table.isIncreaseBlinds()) {
```
ericliu121187, line 53: How can this line be triggered to return `true` for the first hand?

teddylo57: I think line 377 in Table.java has the start new Tournament logic so you would have to set that to true as well

ericliu121187: That was it - also needed to set the blind level to one prior in order to make it compatible.

PR #220, opened 2020-06-29
User does not have lingering hand rank after tournament ends

ericliu121187:
Also, made a style update to replace the underscores with spaces in the hand rank display.

Tested
- Tournament ends and hand ranks disappear https://share.getcloudapp.com/04uYEPEN


backend/src/main/java/poker/services/gameplay/GameService.java
```diff
@@ -42,6 +42,11 @@ public static void startTournament() {
     }
   }
 
+  public static void endTournament() {
+    Api.testTable.getHandRankService().resetHandRanks();
+    Api.testTable.getHandRankService().sendHandRanksToAllPlayers();
```
ericliu121187, line 47: Confirmed this _should_ work for knocked out players given that underneath the hood, this is operating on `playersMap` which should contain all players in the room

PR #222, opened 2020-06-30
User cannot sit or stand when tournament is in progress

ericliu121187:
- User cannot sit or stand when tournament is in progress
- User can sit and stand after a table reset
- Renamed 'Start Tournament' to 'Reset Tournament'
- Handles reconnection

Tested
- User cannot sit or stand when tournament is in progress, can sit and stand after a table reset; renamed 'Start Tournament' to 'Reset Tournament' when a tournament is in progress (https://share.getcloudapp.com/04uYEPEN)
- Reconnection works (https://share.getcloudapp.com/lluJ0DZg)


frontend/src/components/room/index.js
```diff
@@ -86,7 +89,7 @@ function Room({
         </div>
 
         {/* !!userSeat is to satisfy flow */}
-        {isSeated && !!userSeat && (
+        {isSeated && !!userSeat && !isTournamentInProgress && (
```
ericliu121187, line 92: In the future, it might be preferable to have a "disabled" state rather than removing the button entirely from the UI. For now, removal is slightly preferable given the LOE to design a "disabled" button.

PR #223, opened 2020-06-30
Knocked out player are removed from table

teddylo57:
-Issue fixed was that the dealer button was moved to players that have been knocked out and they were not being removed/dealer button could not be set properly


backend/src/main/java/poker/models/game/Table.java
```diff
@@ -69,7 +69,8 @@ private void initializeSeats() {
   }
 
   public void startHand() {
-    setButton();
+    moveButtonToNextDealer();
+    removeEliminatedPlayers();
```
ericliu121187, line 73: This needs to come after `moveButtonToNextDealer`, right? If so, may want to add a comment here or abstract this to a method like `rotateButtonAndRemoveEliminatedPlayers`, since it may not be intuitive?

teddylo57: Sounds good

teddylo57:
They were tested separately, will merge them and test them together

PR #224, opened 2020-06-30
User has correct min raise

ericliu121187:
1. Backend calculates the valid min raise (before taking into chips) and passes it to the frontend.
1. The frontend uses it to calc raise amounts before sending it to the backend.
1. Then, there is one more layer of validation and it will update the the roundMinRaise if needed.


backend/src/main/java/poker/models/game/Table.java
```diff
@@ -45,6 +46,8 @@ public Table(String name, int id, BlindLevel blindLevel) {
     generateNewActionId();
     this.hand = new Hand(this, HandState.NOT_STARTED);
     this.handRankService = new HandRankService(this);
+    this.roundBetAmount = 0;
```
ericliu121187, line 49: Would it make sense to move this to hand?

teddylo57: I'm open to either, hand makes more sense probably

backend/src/main/java/poker/models/game/Hand.java
```diff
@@ -61,7 +62,7 @@ public void dealNewHand(Table table) {
       table.setIncreaseBlinds(false);
     }
     setBlinds(table);
-    this.roundAmountToPlay = blinds.getBigBlind().getAmount();
```
ericliu121187, line 64: Abstracted this away so that only Hand needs to know how to resetRoundValues + there is a single function that handles that

backend/src/main/java/poker/services/gameplay/GameService.java
```diff
@@ -147,10 +147,12 @@ public static void processPlayerAction(Session user, Request request) {
 
       switch (playerAction) {
         case BET:
+          int amount = request.getData().getAmount();
```
ericliu121187, line 150: Did not end up using this - can revert if you prefer

ericliu121187:
@teddylo57 I updated unit tests + here is a screenshot (https://share.getcloudapp.com/Wnub7yD0) of it working with two players. I would prefer it we tested multiway together, going to add a label to remember if that is ok with you.

teddylo57:
Sure lets do a quick round of testing with the new bug fixes

PR #227, opened 2020-07-01
fix null username

teddylo57:


backend/src/main/java/poker/services/users/UserService.java
```diff
@@ -212,6 +214,10 @@ public static boolean isValid(UUID uuid, String username) {
 
   public static List<String> getLoginErrors(UUID uuid, String username) {
     List<String> errors = new ArrayList<>();
+    if (username.length() == 0) {
```
ericliu121187, line 217: What is the problem this is trying to solve, because doesn't [this line](https://github.com/ericliu121187/short-deck-poker/pull/227/files#diff-7cb6c6c9a363aa108bbfd7841ff5f421R222) cover this case? Are we providing a better error message?

teddylo57: I noticed when I accidentally pressed enter on the login, there's a NPE being thrown because username is actually null, so for the case of line 221 when we check if the String username is alpanumeric, it's checking on null

ericliu121187: Should this check for null then? or does `username.length()` work if username is null?

teddylo57: Good point, typically I'd check for null but I don't think the FE sends null, instead just an empty string like 

{"data":{"uid":"69b777f3-9e93-40a7-8cda-ffaffcdbdd20","username":""},"type":"WS_CONNECT_KICKOFF"} and I think that gets deserialized as a null String, but can check it later

PR #228, opened 2020-07-01
Fixes for allIn

teddylo57:
-In multiway, if 2 players go all in and the 3rd and last player folds, it should fast forward to showdown


backend/src/main/java/poker/services/gameplay/GameService.java
```diff
@@ -165,8 +165,12 @@ public static void processPlayerAction(Session user, Request request) {
                   new MiniPlayer(request.getData().getUid(), UserService.getPlayer(user).getUsername()),
                   playerAction);
           UserService.getPlayer(user).fold();
-          nextPlayer.getPlayer().setToAct(true);
-          break;
+          if (nextPlayer.getPlayer().isAllIn()
```
ericliu121187, line 168: It would not hurt to have a comment something like "In multiway, if 2 players go all in and the 3rd and last player folds, it should fast forward to showdown".

backend/src/main/java/poker/services/gameplay/GameService.java
```diff
@@ -165,8 +165,12 @@ public static void processPlayerAction(Session user, Request request) {
...
+                  && Api.testTable.shouldFastForwardToShowdown()) {
+            Api.testTable.getHand().getHandState().performTask(Api.testTable);
+          } else {
+            nextPlayer.getPlayer().setToAct(true);
+          }          break;
```
ericliu121187, line 173: nit: `break;` on new line

PR #229, opened 2020-07-01
User has round bet amounts reset for Showdown

ericliu121187:
When the round bet amounts are reset, it displays the winner and all player bet chip amounts as expected.

Tested
- https://share.getcloudapp.com/12uNzQLy


backend/src/main/java/poker/models/game/HandState.java
```diff
@@ -6,13 +6,10 @@
 import poker.services.chat.ChatService;
 import poker.services.gameplay.GameService;
 import poker.services.gameplay.HandRankService;
-import poker.services.users.UserService;
 
 import java.io.Serializable;
-import java.util.List;
 import java.util.Observable;
 import java.util.Observer;
-import java.util.stream.Collectors;
```
ericliu121187, line 15: Unused.

PR #234, opened 2020-07-02
Fix user can bet a valid amount; add github actions for frontend

ericliu121187:


.github/workflows/frontend.yml
```diff
@@ -0,0 +1,23 @@
+name: Frontend CI
+
+on:
+  push:
+    paths:
+      - '**'
```
ericliu121187, line 6: Would prefer if this ran only when changes are made in /frontend. Tried the following to no avail, will visit later.

`'frontend'`
`'frontend/**'`
`'*/frontend/**'`

